PR #12, opened 2015-02-13
Better support for spaces

wichtounet:
This PR fixes support for options such as --file="folder with spaces/file".

However, I don't think that all the cases are covered, but at least the very simple case of an argument with space is covered. 

What do you think of it ?


jarro2783:
Oh I see, the generic istringstream reader stops at spaces right? I think this works.

wichtounet:
Yes, it separates each field into a different string.

PR #13, opened 2015-03-20
cxxopts in biicode

davidsanfal:
Hi Jarryd,

I am David from biicode, a small start-up that’s developing a C and C++ dependency manager; just like Maven and Maven Central for Java. The tool consists of two parts: the client installed in the user's machine and the cloud, populated with the most popular and demanded libs and frameworks. Some of our users had been requesting for a while already Cxxopts so I decided to upload it and give it full support, and it works like a charm.

My idea is to make a merge request to your main repository to integrate biicode support. I have also included a file, biicode.conf, which allows your library to work with biicode. The only thing I have changed is the CMakeList.txt, but these lines affect only when you want to use biicode, otherwise they are ignored, so you can use Cxxopts with biicode or without it without any problem. I tried to make the small changes it required as crystal clear as possible but feel free to request further more detail.

Now, let me show you how to use **cxxopts** with biicode.

Just clone your repository with the biicode.conf file and go into it. then, run the following commands:

```sh
$ git clone https://github.com/davidsanfal/cxxopts.git
$ cd cxxopts
$ bii init -L
$ bii build
```

That's all, biicode will have compiled your code into the bin folder. To run it just execute:

```sh
$ cd bin
$ ./*_cxxopts_src_example -abc 3
Saw option ‘a’
Saw option ‘b’
Positional = 3
Output = a.out
Arguments remain = 1
```

I have published this library on biicode, in [david/cxxopts](https://www.biicode.com/david/cxxopts).

We would love you to sign up in biicode so that you could maintain and keep your library up-to-date in biicode yourself. Our users would like to have their code dependent on the latest stable version of your code and there's no one better than you to keep it alike, beside that would allow any of our users to reuse your lib in any kind of project with just one `#include`.

Plus we have a [smooth workflow with GitHub and Travis CI](http://blog.biicode.com/automatically-build-publish-via-travis-ci-github/).

Looking forward to your reply.

Best regards.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,6 +18,14 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 
+IF(BIICODE)
+	include(biicode/cmake/tools)
+	ADD_BIICODE_TARGETS()
+	ACTIVATE_CPP11(INTERFACE ${BII_BLOCK_TARGET})
+RETURN()
+
+ENDIF()
+
 cmake_minimum_required(VERSION 2.6)
 project(cxxopts)
 
@@ -37,6 +45,3 @@ if(CXXOPTS_USE_UNICODE_HELP)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DCXXOPTS_USE_UNICODE")
 
 endif()
-
-add_subdirectory(src)
-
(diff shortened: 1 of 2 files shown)


jarro2783:
This seems reasonable. Would it be easier if you continue to use the latest code from github, rather than keeping two copies synchronised?

davidsanfal:
Hi. In fact, if you accept this merge request you can use your own repo, in this example I use my repository because I understand that you prefer to try it before you accept.

Also you can automate the publication to biicode using a CI like travis.

PR #23, opened 2016-05-23
Add MSVC 2013 compiler support

nocnokneo:


src/cxxopts.hpp
```diff
@@ -38,6 +38,7 @@ THE SOFTWARE.
 #include <sstream>
 #include <string>
 #include <vector>
+#include <string.h>
```
Nava2: This should be `#include <cstring>`.

Nava2:
I'm having a problem with the `strcmp` issue on clang 3.7, not just MSVC 2013.

jarro2783:
I've left this one for a while because I've had to think about it a bit.  It's unfortunate that this code doesn't work with the microsoft compiler, so it should be fixed. However, this patch does not match my vision for this library, so at the moment I'm going to reject it.

I want this to be a simple library that can be included with no fuss. The cmake file is there to compile the example, it is not intended to be used by users of the library. Adding a dependency header that can only be generated by cmake means that anyone using the library needs to get that same dependency header into their build system.

I also don't understand why MSVC doesn't support unicode literals. Is this an old version that doesn't have complete C++11 support?

I might be open to a patch so that the header still works standalone.

I will fix the cstring issue.

Nava2:
Thanks for adjusting the cstring issue.

nocnokneo:
The best reference I'm aware of for which compilers support which features is from cppreference.com:

http://en.cppreference.com/w/cpp/compiler_support

Note that MSVC 14.0 is what comes with Visual Studio 2015 and MSVC 12.0 is what comes with Visual Studio 2013 - so just one version behind the newest available from Microsoft.

I agree with the vision of a single self-contained .hpp file that can easily drop into any project. Checking for the value of `_MSC_VER` should be an easy way to add the special case macro definitions needed for MSVC 12.0:

http://stackoverflow.com/a/70630

I'll give this a try and re-submit.

PR #24, opened 2016-05-23
Don't count default values

nocnokneo:
This makes option parsing of something like this possible:

```cpp
  m_options.add_options()
    ("output-file", "Output filename", cxxopts::value(m_outputFile))
    ("output-format", "Output format", cxxopts::value(m_outputFormat)->default_value("csv"))
    ;

  m_options.parse(argc, argv);
  if (!m_options.count("metadata-format"))
  {
    // output format not specified explicitly - try to figure it out
    // implicitly by m_outputFile extension
    // ...
  }
```

### DIFF
--- a/src/cxxopts.hpp
+++ b/src/cxxopts.hpp
@@ -609,7 +609,6 @@ namespace cxxopts
     parse_default()
     {
       m_value->parse();
-      ++m_count;
     }
 
     int


jarro2783:
I don't understand what you're trying to do here. What's the point of having a default value if you're going to ignore it anyway?

nocnokneo:
Maybe it's easier to show by usage examples:

```bash
myapp --output-file foo.xml # writes foo.xml in xml format (determined by file extension)
myapp --output-file foo.txt --output-format xml # writes foo.txt in xml format
myapp --output-file foo # writes foo in csv format (determined by default value for --output-format)
myapp --output-file - # writes to stdout in csv format (determined by default value for --output-format)
```

jarro2783:
I can see what you're getting at, and it seems reasonable. My only concern is that it might break existing code relying on defaults being counted. Not being able to survey everyone using this, I have no idea what might happen. I think I'll accept this now, but if people complain it might have to go back, or be an option rather than the default.

PR #29, opened 2016-06-18
Visual improvements

pmed:
Hi,

Thanks for the really lightweight library. I just wanted to share a couple of improvements for options help output.

### DIFF
--- a/src/cxxopts.hpp
+++ b/src/cxxopts.hpp
@@ -884,7 +884,7 @@ namespace cxxopts
 
       if (o.has_default)
       {
-        desc += toLocalString(" (default:" + o.default_value + ")");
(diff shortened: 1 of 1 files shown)


jarro2783:
Can you split this pull request? I like the look of the newline skipping in the help, but there is something funny with the characters in that commit.

I think I will do something different for the quotes, since they should probably still be included when CXXOPTS_USE_UNICODE is defined.

pmed:
I've extracted only changes for options group help.

Don't know how those changes related to UTF-8 appeared in the previous commit :confused:  However, I've saved them in a branch [pmed/cxxopts:no-utf8-literals](https://github.com/pmed/cxxopts/tree/no-utf8-literals).

PR #33, opened 2016-11-29
Use cmake interface library

mariobadr:
This PR uses cmake's interface library feature:

> An INTERFACE library target does not directly create build output, though it may have properties set on it and it may be installed, exported and imported. 

This makes it easier to include the header only library in a cmake project. After using add_subdirectory on the cxxopts directory, one simply needs to include `cxxopts` in their target_link_libraries, which will allow the user's target to inherit the properties of the cxxopts header library (see changes to example and test).

Two things: 

(1) I updated the cmake required version to 3.1, but this may not be necessary (I believe it is necessary for target_sources, which is not available in 2.8.12, but not sure if you need target_sources for this to work). 

(2) The Unicode support is confusing. I did a git grep and couldn't find any other use for CXXOPTS_LINKER_LIBRARIES. CXXOPTS_USE_UNICODE_HELP would also make more sense as an option instead of a cached variable, and then target_compile_definitions can be used to enable CXXOPTS_USE_UNICODE in the header. With this PR making it an interface library, other projects that link with cxxopts should inheret CXXOPTS_USE_UNICODE as well. For now I left this untouched since I don't know if I'm missing something important.

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/33)


jarro2783:
This looks like a good change in general, but see comments below.

Updating the minimum cmake version is fine.

I think the Unicode support is broken at the moment, feel free to fix it too. What should probably happen is that you ask for Unicode, and anyone who uses the library should get both `-DCXXOPTS_USE_UNICODE`, and any required link libraries. See 8d7c4ea43e25a3187662506040b826a167e7110c, and then 774d455dc720dc0d90d4d0bbdc57bde0ccb9f51d which I think broke linking.

---

Reviewed 7 of 7 files at r1.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[CMakeLists.txt, line 51 at r1](https://reviewable.io:443/reviews/jarro2783/cxxopts/33#-KXlsVAACDeE6uEoHQsR:-KXlsVAB2lfL5w0ZYcBQ:b-23adza) ([raw file](https://github.com/jarro2783/cxxopts/blob/151c72382b131aa5b03aa194bae4f1295a8a79b0/CMakeLists.txt#L51)):*
> ```CMake
> endif()
> 
> add_library(cxxopts INTERFACE)
> ```

Is there any reason that you can't include the sources in `add_library`, that way you won't need `target_sources`?

---

*[CMakeLists.txt, line 54 at r1](https://reviewable.io:443/reviews/jarro2783/cxxopts/33#-KXlt7ybwU1Br7FQP54G:-KXlt7ybwU1Br7FQP54H:bnk6zhd) ([raw file](https://github.com/jarro2783/cxxopts/blob/151c72382b131aa5b03aa194bae4f1295a8a79b0/CMakeLists.txt#L54)):*
> ```CMake
> target_sources(
>   cxxopts INTERFACE
>   ${CMAKE_CURRENT_SOURCE_DIR}/include/cxxopts/cxxopts.hpp
> ```

I'm ok with moving the headers to `include`, because reorganising the source has no external effect, and to be consistent with the rest of the world it's probably the right thing.
But I don't think there's any point putting `cxxopts.hpp` under `cxxopts`, because it is a single file. If there were multiple files I think it would make sense for them to live in their own directory. For example we have `/usr/include/gmp.h`, but boost has its own directory. This also breaks backwards compatibility.

---

*[src/CMakeLists.txt, line 23 at r1](https://reviewable.io:443/reviews/jarro2783/cxxopts/33#-KXlvEZnUJ8m_zcldx_P:-KXlvEZnUJ8m_zcldx_Q:b3kdijb) ([raw file](https://github.com/jarro2783/cxxopts/blob/151c72382b131aa5b03aa194bae4f1295a8a79b0/src/CMakeLists.txt#L23)):*
> ```CMake
> if(CXXOPTS_BUILD_EXAMPLES)
>     add_executable(example example.cpp)
>     target_link_libraries(example PRIVATE cxxopts)
> ```

What does `PRIVATE` do here? I don't understand the docs on this one.

---

*[src/example.cpp, line 27 at r1](https://reviewable.io:443/reviews/jarro2783/cxxopts/33#-KXlvQlHSwtuZC9WvTOu:-KXlvQlHSwtuZC9WvTOv:b-nrvimj) ([raw file](https://github.com/jarro2783/cxxopts/blob/151c72382b131aa5b03aa194bae4f1295a8a79b0/src/example.cpp#L27)):*
> ```cpp
> #include <iostream>
> 
> #include "cxxopts/cxxopts.hpp"
> ```

As mentioned above I don't think we should be changing the relative include path.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jarro2783/cxxopts/33#-:-KXlvvvVDLuZXTCp-egV:b-jir0d8)*

jarro2783:
I've fixed the Unicode stuff. Can you take a look at a04284fc2c650f52e0d09cbc6c6d688fbf64cca9 see if that affects this PR? I suspect that yours is the right behaviour, but at least now it goes from one working setup to another, instead of being unsure about something that's broken.

---

Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/jarro2783/cxxopts/33#-:-KXo2gyO0U54RQBWEi_e:b-yojetp)*

mariobadr:
Couldn't figure out if there was a way to respond to your comments on reviewable, so please find some responses here:

mariobadr:
Forgot that I removed `install(FILES cxxopts.hpp DESTINATION include)` from `src/CMakeLists.txt`. Would you like me to add it back to the root CMakeLists.txt?

jarro2783:
You should be able to sign in to reviewable with github and reply to the comments there.

Can you rebase on top of master and force push instead of pulling my new commits into your branch? I can't tell what has changed now, and it has also messed up what reviewable thinks of the diff. Plus there are merge conflicts that I don't think should actually be there.

> Forgot that I removed install(FILES cxxopts.hpp DESTINATION include) from src/CMakeLists.txt. Would you like me to add it back to the root CMakeLists.txt?

Yes that would be good.

> It shouldn't affect the PR, I don't think, but here's where INTERFACE/PUBLIC/PRIVATE come in. If I target_link_library with cxxopts (instead of example, as you're doing), and specify PUBLIC, this means that the library is linked with cxxopts and any executables that link with cxxopts. If I specify INTERFACE, this means that cxxopts is not linked with CXXOPTS_LINKER_LIBRARIES, however executables that link to cxxopts will link with CXXOPTS_LINKER_LIBRARIES.

This sounds like the right thing to do then.

jarro2783:
Review status: 0 of 6 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[src/CMakeLists.txt, line 31 at r2](https://reviewable.io:443/reviews/jarro2783/cxxopts/33#-KXrKz4Fx7MyIOfr2V6y:-KXrKz4Fx7MyIOfr2V6z:bpqv0p2) ([raw file](https://github.com/jarro2783/cxxopts/blob/d92988c6a218e2fef987e30d99445e7c87c4b87b/src/CMakeLists.txt#L31)):*
> ```CMake
> endif()
> 
> install(FILES cxxopts.hpp DESTINATION include)
> ```

I missed this, can you put it in the root file.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jarro2783/cxxopts/33)*

mariobadr:
> Can you rebase on top of master and force push instead of pulling my new commits into your branch?

Sorry, can you help with the git commands? I'm still learning...

mariobadr:
I think I got it, I did `git rebase master && git push -f`, let me know if I messed anything up. I can redo my changes in a new pull request if you'd prefer.

jarro2783:
:lgtm:

---

Reviewed 1 of 8 files at r1, 6 of 8 files at r2, 1 of 1 files at r4.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/jarro2783/cxxopts/33#-:-KXrgv-I8kvrkd8p5E6R:bnfp4nl)*

jarro2783:
I'll merge this as soon as I fix my travis build.

Issue #34, opened 2016-12-15
error parsing options: Option ├óΓé¼╦£t├óΓé¼Γäó does not exist


mulderp:
I get a strange strings when running this code without a valid option:

 
```
 options.add_options("")
	("a,aaaa", "aaaa type", cxxopts::value<std::string>()->default_value("bbbb"), "bbbb or cccc")
			("h,help", "Print help");
	options.parse(argc, argv);
	if (options.count("help"))
	{
		std::cout << options.help({ "", "Group" }) << std::endl;
		exit(0);
	}


```
 

Build is on Windows with MSVC community 15

petewall:
Same here.  When options.parse encounters an option that it doesn't recognize, it's throwing an option_not_exists_exception.  The output of that exception (using `what()`) has the same text no matter which unknown argument I provide.

I'm on Visual Studio Basic 2015 on Windows 10 x64

jarro2783:
Well that's different. Technically the output is not lying, but it's probably also not what you want :stuck_out_tongue: 
I'll take a look.

jarro2783:
Well this is going to make life difficult. I can't reproduce this on Linux, and I don't have access to a Windows machine.

Do you have a tool like Valgrind that you can run it with? That or if you're able to run it in a debugger and provide me with some more output then that will help me track this down.

petewall:
I can do some debugging later (maybe tomorrow).

petewall:
Ok, so actually, the option name isn't being trashed, but instead, it's the and the beginning of the string.  Notice after the "£" is the character "x", which is the option I used to test with:
`Option ├óΓé¼╦£x├óΓé¼Γäó does not exist`

@mulderp 's message (`Option ├óΓé¼╦£t├óΓé¼Γäó `) shows a "t", so he must have been testing with the character "t".

Looking at how that string is constructed, we see this:
```
class option_not_exists_exception : public OptionParseException
  {
    public:
    option_not_exists_exception(const std::string& option)
    : OptionParseException(u8"Option ‘" + option + u8"’ does not exist")
    {
    }
  };
```

if I step in, the debugger goes the operator+ function for the string.  The left operand is "Option Ã¢â‚¬Ëœ" and the right operand is "x".

This makes me think that Windows, VS2015, or both, do not like the single quote's that are a part of the exception string.  After replacing the `‘` with `'`, it works:

`Option 'x' does not exist`

I'll make the changes and submit a pull request.

[...]

PR #35, opened 2016-12-20
Replace curly single-quotes with a standard quote character.  

petewall:
This fixes an issue with printing those characters on Windows.

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/35)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -318,7 +318,7 @@ namespace cxxopts
   {
     public:
     option_exists_error(const std::string& option)
-    : OptionSpecException(u8"Option ‘" + option + u8"’ already exists")
+    : OptionSpecException(u8"Option '" + option + u8"' already exists")
     {
     }
   };
@@ -327,7 +327,7 @@ namespace cxxopts
   {
     public:
     invalid_option_format_error(const std::string& format)
-    : OptionSpecException(u8"Invalid option format ‘" + format + u8"’")
+    : OptionSpecException(u8"Invalid option format '" + format + u8"'")
     {
     }
   };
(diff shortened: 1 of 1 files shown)


jarro2783:
As mentioned in #34, I don't want to throw away quotes just because Windows is deficient. I would prefer that single quotes be used only if on a platform that can't handle UTF-8, overridable by the user requesting curly quotes explicitly.

jarro2783:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[include/cxxopts.hpp, line 321 at r1](https://reviewable.io:443/reviews/jarro2783/cxxopts/35#-KZT5eRoHAHNYjsjqVb9:-KZT5eRoHAHNYjsjqVbA:bsf2rat) ([raw file](https://github.com/jarro2783/cxxopts/blob/fe78ff39fbb26ec248a7daf1ab3e3fb9d6babb4b/include/cxxopts.hpp#L321)):*
> ```cpp
>     public:
>     option_exists_error(const std::string& option)
>     : OptionSpecException(u8"Option '" + option + u8"' already exists")
> ```

I would prefer to make these conditional on platforms that aren't sane.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jarro2783/cxxopts/35)*

jarro2783:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/jarro2783/cxxopts/35)*

Anton-V-K:
Any chances these changes will soon be merged into the base branch?

jarro2783:
As I said in my comment, I like the curly quotes, as do a lot of unix tools (GCC). I would be happy to disable them with an `#ifdef`.

jarro2783:
Fixed by #68.

Issue #39, opened 2017-02-15
Unsigned parameters accept negative arguments


RedDwarf69:
istringstream has a "problem" -> http://stackoverflow.com/questions/18917167/stringstream-unsigned-input-validation

I would argue that cxxopts should reject negative arguments when the parameter type is unsigned.

jarro2783:
Yes I think I agree with that.

PR #40, opened 2017-03-09
Interpret input starting with 0x as hex numbers.

da1l6:
When giving numeric values are hex (0xF00) cxxopts will currently refuse these.
This patch will check for 0x prefix and parse the value as hex.

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/40)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -404,9 +404,17 @@ namespace cxxopts
     parse_value(const std::string& text, T& value)
     {
       std::istringstream is(text);
-      if (!(is >> value))
-      {
-        throw argument_incorrect_type(text);
+      if (text.substr(0, 2) == "0x"){ //Hex number input
+        is.seekg(2);
+        if (!(is >> std::hex >> value))
+        {
+           throw argument_incorrect_type(text);
+        }
+      } else {
+        if (!(is >> value))
+        {
+           throw argument_incorrect_type(text);
+        }
       }
 
       if (is.rdbuf()->in_avail() != 0)


jarro2783:
I think I would prefer to parse numbers properly instead of doing this hack in the main parser. #39 also shows that numbers are not parsed correctly by `istringstream`.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/jarro2783/cxxopts/40#-:-Kj4zplVuGVcTDvBzAT4:bqtolud)*

niosHD:
I recently also had the need for passing a `uint64_t` as a hex numbers via the command line. My workaround was to take a `std::string` and do the parsing manually in the code with [std::stoull](http://en.cppreference.com/w/cpp/string/basic_string/stoul). Even though it is not much overhead in terms of additional code, I also think that it would be a nice feature for cxxopts to automatically accept different bases for integers.

Interestingly, the `std::std...` conversion functions already support automatic base detection. I therefore thought about specializing `cxxopts::values::parse_value` for the common integer types and using the standard conversion functions to get the desired behavior. Unfortunately, this approach would not address #39 since the standard conversion functions accept a unary minus in front of unsigned types as well.

Would this be a better approach to the problem or are there other suggestions? I could give it a try if it is acceptable.

jarro2783:
I plan on rewriting this part of the code so that it works properly for all types of integers.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/jarro2783/cxxopts/40#-:-Km-e2cZRbJd34yqee3K:b-g2os17)*

Digicrat:
@jarro2783 Any updates?  This is the first real issue I've encountered since starting to use this convenient lightweight library a few weeks ago.

PR #51, opened 2017-06-05
Added inline to check_required

Domo929:
This alleviated an issue I had where during compilation, check_required was defined multiple times. It looked like the only function that did not have the inline modifier before it, which explains why it was the only one to be defined multiple times.

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/51)


jarro2783:
Ah, oops, thanks.

Issue #52, opened 2017-06-12
no cxxoptsConfig.cmake after installing with cmake ../ && make install


Glavnokoman:

jarro2783:
I don't know what this file is supposed to be. Why would I expect it to exist?

Glavnokoman:
This is to make it easy to use cxxopts with cmake projects. For those who
do not want to include it in the project sources but to access through
findPackage.

jarro2783:
#53 has given me some more context. I will get to that pull request and that should sort out the problem.

jarro2783:
Closed by #53

PR #53, opened 2017-06-19
Cmake overhaul

niosHD:
Hi,

this PR is a overhaul of the build system with the goal of adding the CMake config files which have been requested in #52. These files are used by CMake's `find_package()` command to import the library target into a consuming project when cxxopts has, for example, been installed globally. Since cxxopts is a header only library it is of course not really necessary to use the `find_package` machinery. However, I think it is quite convenient for the users to provide the standard cmake way of consuming external libraries. The two tests show how cxxopts can be used with `find_package` and `add_subdirectory` respectively.

I also took the liberty to remove the biicode support since the project is dead for quite some time already.

Regards,
Mario


jarro2783:
Thanks for this, a quick glance looks like it is a Good Thing. I will have a closer look at this a bit later when I have some time. I don't know CMake that well, so what I currently have is *something* that works, rather than any best practices.

---

Review status: 0 of 7 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/jarro2783/cxxopts/53#-:-Kn1S-zzpWwhqVVB9r-H:b-y6a28o)*

jarro2783:
Looks good. Just a couple of questions so that I understand what's going on.

---

Reviewed 7 of 7 files at r1.

---

*[test/add-subdirectory-test/CMakeLists.txt, line 8 at r1](https://reviewable.io:443/reviews/jarro2783/cxxopts/53#-KnGrcBuyw8ArAyJ31CO:-KnGrcBuyw8ArAyJ31CP:b-eurkta) ([raw file](https://github.com/jarro2783/cxxopts/blob/8a92fa2bffa3e3574161ee5f611d6c5851c2240b/test/add-subdirectory-test/CMakeLists.txt#L8)):*
> ```CMake
> set(CMAKE_CXX_EXTENSIONS OFF)
> 
> add_subdirectory(../.. cxxopts EXCLUDE_FROM_ALL)
> ```

What does this do?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jarro2783/cxxopts/53#-:-KnGuZkISrscVJ908dmv:bsudesg)*

niosHD:
Sure, ask anything you want to know.

---

---

*[test/add-subdirectory-test/CMakeLists.txt, line 8 at r1](https://reviewable.io:443/reviews/jarro2783/cxxopts/53#-KnGrcBuyw8ArAyJ31CO:-KnIkqFTydafNF3jBxTQ:bkc2cxp) ([raw file](https://github.com/jarro2783/cxxopts/blob/8a92fa2bffa3e3574161ee5f611d6c5851c2240b/test/add-subdirectory-test/CMakeLists.txt#L8)):*
<details><summary><i>Previously, jarro2783 wrote…</i></summary><blockquote>

What does this do?
</blockquote></details>

The idea of the test is to make sure that a consumer of the library can simply add the whole cxxopts directory as subdirectory into his project. This is useful when a user for example has cxxopts as git submodule in his own project. If everything works as intended then the user can simply link against the library.

The specific line does exactly that. It simply adds the cxxopts root directory as subdirectory to the test project. Since in this specific case the added directory is not really a subdirectory, cmake additionally requires a name which will be used in the build tree. Typically, one would simply use `add_subdirectory(cxxopts)`. Additionally, adding `EXCLUDE_FROM_ALL` makes sure that no target in the cxxopts directory (e.g., `example`) is built by default (i.e., with `all` or `install`) but only when they are really needed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jarro2783/cxxopts/53#-:-KnIpw3iQkA6eP894rv4:b-j7izng)*

jarro2783:

---

*[test/add-subdirectory-test/CMakeLists.txt, line 8 at r1](https://reviewable.io:443/reviews/jarro2783/cxxopts/53#-KnGrcBuyw8ArAyJ31CO:-KnJ2m8Wbl7W2QZO9Bcc:biah5lo) ([raw file](https://github.com/jarro2783/cxxopts/blob/8a92fa2bffa3e3574161ee5f611d6c5851c2240b/test/add-subdirectory-test/CMakeLists.txt#L8)):*
<details><summary><i>Previously, niosHD (Mario Werner) wrote…</i></summary><blockquote>

The idea of the test is to make sure that a consumer of the library can simply add the whole cxxopts directory as subdirectory into his project. This is useful when a user for example has cxxopts as git submodule in his own project. If everything works as intended then the user can simply link against the library.

The specific line does exactly that. It simply adds the cxxopts root directory as subdirectory to the test project. Since in this specific case the added directory is not really a subdirectory, cmake additionally requires a name which will be used in the build tree. Typically, one would simply use `add_subdirectory(cxxopts)`. Additionally, adding `EXCLUDE_FROM_ALL` makes sure that no target in the cxxopts directory (e.g., `example`) is built by default (i.e., with `all` or `install`) but only when they are really needed.
</blockquote></details>

So in the *built* tree, you're checking that you can link against the `cxxopts` library by building something that uses it? The example code being a good candidate instead of adding some other small file that uses the library.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jarro2783/cxxopts/53)*

jarro2783:
:lgtm:

---

Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/jarro2783/cxxopts/53#-:-KnJ57l_rsuzJ1JiJtIK:bnfp4nl)*

niosHD:
*[test/add-subdirectory-test/CMakeLists.txt, line 8 at r1](https://reviewable.io:443/reviews/jarro2783/cxxopts/53#-KnGrcBuyw8ArAyJ31CO:-KnJZ-uCk-kf4XZplRQN:b-i13pay) ([raw file](https://github.com/jarro2783/cxxopts/blob/8a92fa2bffa3e3574161ee5f611d6c5851c2240b/test/add-subdirectory-test/CMakeLists.txt#L8)):*
<details><summary><i>Previously, jarro2783 wrote…</i></summary><blockquote>

So in the *built* tree, you're checking that you can link against the `cxxopts` library by building something that uses it? The example code being a good candidate instead of adding some other small file that uses the library.
</blockquote></details>

Yes, exactly. The `add-subdirectory-test` tries to use cxxopts by simply embedding the complete directory into the build system of the consuming project. The `find-package-test`on the other hand tries to use cxxopts using cmake's `find_package` functionality which relies on the (not installed versions of the) files requested in #52.  Both tests, simply reuse the example source file instead of adding new test programs.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jarro2783/cxxopts/53)*

Issue #60, opened 2017-08-05
Parsing types other than standard ones


dagenest:
README.md says "_Any type can be given as long as it can be parsed, with operator>>_." but it does not work for types such as enum classes. In fact, i don't see any call to operator>> in the source.
Defining void parse_value(const std::string& text, _MYTYPE_ & value) works perfectly, and throwing a cxxopts::argument_incorrect_type in parse_value for a wrong value also works perfectly.
Is defining parse_value the right way to do it (so i suppose the README needs an update, because it is a interesting feature), or is it just an undocumented feature that could be removed in future releases?

jarro2783:
Ah right, I think I actually broke that recently with the improved integer parsing.

From 5e9f8225982e9711b548c886fe6b477cdefd0fd4:

```diff
     template <typename T>
     void
-    parse_value(const std::string& text, T& value)
+    integer_parser(const std::string& text, T& value)
     {
-      std::istringstream is(text);
-      if (!(is >> value))
```

So I actually did previously intend to use `operator>>` to parse values. I will put that back in.

Of course you can override it by overloading `parse_value`, and that would be the correct way to work around it for now.

PR #63, opened 2017-08-11
Added parsing of floating point number arguments.

hrittich:
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/63)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -533,6 +533,16 @@ namespace cxxopts
       }
     }
 
+    template <typename T>
+    void stringstream_parser(const std::string& text, T& value)
+    {
+      std::stringstream in(text);
+      in >> value;
+      if (!in) {
+        throw argument_incorrect_type(text);
+      }
+    }
+
     inline
     void
     parse_value(const std::string& text, uint8_t& value)
@@ -603,6 +613,15 @@ namespace cxxopts
     parse_value(const std::string& text, std::string& value)
     {
       value = text;
+    }
+
+    // The fallback parser. It uses the stringstream parser to parse all types
+    // that have not been overloaded explicitly.  It has to be placed in the
+    // source code before all other more specialized templates.
+    template <typename T>
+    void
+    parse_value(const std::string& text, T& value) {
+      stringstream_parser(text, value);
     }
 
     template <typename T>
--- a/src/example.cpp
+++ b/src/example.cpp
@@ -49,6 +49,7 @@ int main(int argc, char* argv[])
         "thisisareallylongwordthattakesupthewholelineandcannotbebrokenataspace")
       ("help", "Print help")
       ("int", "An integer", cxxopts::value<int>(), "N")
+      ("float", "A floating point number", cxxopts::value<float>())
       ("option_that_is_too_long_for_the_help", "A very long option")
     #ifdef CXXOPTS_USE_UNICODE
       ("unicode", u8"A help option with non-ascii: à. Here the size of the"
@@ -116,6 +117,11 @@ int main(int argc, char* argv[])
     if (options.count("int"))
     {
       std::cout << "int = " << options["int"].as<int>() << std::endl;
+    }
+
+    if (options.count("float"))
+    {
+      std::cout << "float = " << options["float"].as<float>() << std::endl;
     }
 
     std::cout << "Arguments remain = " << argc << std::endl;
--- a/test/options.cpp
+++ b/test/options.cpp
@@ -336,3 +336,31 @@ TEST_CASE("Integer overflow", "[options]")
   options.parse_positional("positional");
   CHECK_THROWS_AS(options.parse(argc, argv), cxxopts::argument_incorrect_type);
 }
+
+TEST_CASE("Floats", "[options]")
+{
+  cxxopts::Options options("parses_floats", "parses floats correctly");
+  options.add_options()
+    ("double", "Double precision", cxxopts::value<double>())
+    ("positional", "Floats", cxxopts::value<std::vector<float>>());
+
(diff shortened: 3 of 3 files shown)


include/cxxopts.hpp
```diff
@@ -600,6 +610,18 @@ namespace cxxopts
 
     inline
     void
+    parse_value(const std::string& text, float& value) {
+      stringstream_parser(text, value);
+    }
+
+    inline
+	void
+	parse_value(const std::string& text, double& value) {
```
jarro2783: Remove the extra indents here.

hrittich: Done.

jarro2783:
I'm sure I left a comment that seems to have been swallowed up somehow. See #60, can you make `stringstream_parser` the default for all `<T>`?

hrittich:
The code should now work with types that overload >> .

patrikhuber:
@jarro2783 I seem to be unable to use _any_ floating point values as command-line arguments, i.e.

```
float value;
("f,float", "A float", cxxopts::value<float>(value))
``` 

does not compile with compile error

> Severity	Code	Description	Project	File	Line	Suppression State
Error	C2665	'cxxopts::values::parse_value': none of the 10 overloads could convert all the argument types	test	\cxxopts\include\cxxopts.hpp	671	

Is that supposed to be the case without this PR?

Actually, I want to use a vector<float> as command-line argument. I see from the documentation that vector<string> works out-of-the-box, but not float. Is that supposed to be the case, and will this PR fix that?

Thank you!

hrittich:
@patrikhuber Please see #60.

patrikhuber:
@hrittich I read #60 already. It doesn't specifically talk about whether float parsing (not vector<float> but simple float) should be working right now (without this PR).

I'm sorry if I missed something but the documentation isn't very clear and `int` parsing currently seems to work so it's odd to me why `float` does not.

hrittich:
@patrikhuber cxxopts contains a specialized parser for integers. This allows, e.g., to specify numbers in hexadecimal form, when they start with "0x". This specialized number parser, however, does not parse floating point numbers, which is the reason for this pull request.

patrikhuber:
@hrittich I see! That makes complete sense. Thank you very much for the explanation! :-)

patrikhuber:
I'm not sure whether this is working just yet. How can I make a `vector<float>` option work? I couldn't find anything in the documentation about `vector<T>` as options.

I tried:
`--myoption 1.0 0.0 -1.0`
`--myoption=1.0 0.0 -1.0`
`--myoption=1.0,0.0,-1.0`
`--myoption="1.0 0.0 -1.0"`
`--myoption="1.0,0.0,-1.0"`

I'm not sure what else to try?

hrittich:
The following should work. You need to repeat the option name.
`--myoption 1.0 --myoption 0.0 --myoption -1.0`

patrikhuber:
Oh wow, that's horribly ugly and inconvenient. Not really useful for my use case in fact.
Is there a way to change that into a more useful syntax without repeating the option name?

hrittich:
The reason why cxxopts allows to use vectors for argument storage is to allow users to specify an argument more than once and the program to have access to all those arguments. The idea is not to actually have the user input something like a "logical" vector.

I do not see a suitable syntax that is easy to implement for your use case. For example, using a comma as a delimiter might fail for strings. The same is true for using spaces as delimiters. Someone would have to implement some rather complex parser. It would be possible, however, much work.

I would recommend you to parse the option as a string and then split it, the way you like.

jarro2783:
Thanks @hrittich. Yes everything you said is correct.

You could also use a vector to hold the last positional argument, and then every positional argument that didn't go into another option would be added onto the end of that one.

The reason I made it like that is because a lot of programs like to allow an option to be written multiple times, so this is how you differentiate zero or one allowed, or more than one allowed.

If you wanted to parse your own you could set your value to be of some type `T`, and then provide an `operator>>(std::istream&, T&)` function.

patrikhuber:
Thanks for the hints to both of you!

I tried the approach with a custom type but I did not succeed:

```
struct Coefficients
{
    std::vector<float> coefficients;
};


std::istream& operator>>(std::istream& is, Coefficients& c)
{
    double test;
    is >> test;
    c.coefficients.push_back(test);
    return is;
};
```
Then adding an option: `("shape-coeffs", "Docs", cxxopts::value<Coefficients>(coefficients))`
But if I now start my program with ` --shape-coeffs 1.0 2.0`, the `operator>>` overload only gets called once, for the `1.0`, and the `2.0` seems to get swallowed.

hrittich:
The operator is only called whenever it encounters the option "--shape-coeffs". Then, you have to make sure that you read all the arguments that you think belong to the option, meaning, you somehow need to know how many numbers you have to read.

patrikhuber:
I guess I could read up to the next long option (if there is any), that would work. But I have no idea from the cxxopts code I've seen so far how to do that. I think I may just go back to boost program_options unfortunately :-\ Or is that easy to accomplish?

jarro2783:
There is no way for the parser to know that `--shape-coeffs` can take multiple arguments. You would have to just write it down multiple times. The other option would be to define your own type with an `operator>>`, and have the user write it as something like `1.0,2.0`.

patrikhuber:
Yea I tried using my own type and define `operator>>`, but as you say, that would only work if there is no space between the numbers (like in `1.0,2.0`). I do not think it's a good option to put this burden on the user to don't allow white-spaces there, many will put white-space there, even if you clearly indicate otherwise in the documentation (I guess you know how "general" users are, too ;-) ).
I went back to boost program_options now. If cxxopts ever has facilities to accomplish this (and I don't think it's a too far-fetched use-case to be honest), I would be looking forward very much to switching back to it!

PR #68, opened 2017-08-28
Remove curly single-quotes only on Windows

giumas:
https://github.com/jarro2783/cxxopts/pull/35

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/68)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -243,6 +243,14 @@ namespace cxxopts
 
 namespace cxxopts
 {
+#ifdef _WIN32
+  const std::string LQUOTE("\'");
+  const std::string RQUOTE("\'");
+#else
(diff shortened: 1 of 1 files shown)


jarro2783:
Thanks for the pull request. My only comment is that this would probably be cleaner doing something like:

```
#ifdef _WIN32
  #define LQUOTE "'"
  #define RQUOTE "'"
#else
  #define LQUOTE "’"
  #define RQUOTE "’"
#endif
```

and then use `LQUOTE` and `RQUOTE` as appropriate. That way the logic can be enhanced without changing a whole bunch of code.

giumas:
@jarro2783 I prefer a `const std::string` solution, but feel free to modify my PR as you like.

jarro2783:
That works too.

PR #70, opened 2017-08-30
Fixup integer pattern multiple definition

### LINKED ISSUE #69: Multiple definitions of `cxxopts::values::integer_pattern`
Creating a global value does makes one version of it per compilation unit, which causes problems at link time when cxxopts is used in more than one compilation units.

Instead of doing

```cpp
namespace
{
  std::basic_regex<char> integer_pattern
    ("(-)?(0x)?([1-9a-zA-Z][0-9a-zA-Z]*)|(0)");
}
```

you probably should have

```cpp
inline const std::basic_regex<char>& integer_pattern()
{
    static const std::basic_regex<char> pattern = "(-)?(0x)?([1-9a-zA-Z][0-9a-zA-Z]*)|(0)";
    return pattern;
}
```

i.e. make a singleton.

SylvainCorlay:
Fixes #69 

Since the value was only used once, I did not need to make a singleton as initially proposed in #69.

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/70)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -428,11 +428,6 @@ namespace cxxopts
 
   namespace values
   {
-    namespace
-    {
-      std::basic_regex<char> integer_pattern
-        ("(-)?(0x)?([1-9a-zA-Z][0-9a-zA-Z]*)|(0)");
-    }
 
     namespace detail
     {
@@ -484,6 +479,8 @@ namespace cxxopts
     integer_parser(const std::string& text, T& value)
     {
       std::smatch match;
+      std::basic_regex<char> integer_pattern
+        ("(-)?(0x)?([1-9a-zA-Z][0-9a-zA-Z]*)|(0)");
       std::regex_match(text, match, integer_pattern);
 
       if (match.length() == 0)


SylvainCorlay:
I confirm that this fixes our build. If you really like to make it global, we can use a singleton as described in the issue.

SylvainCorlay:
Are you ok with merging this? FYI, the conda recipe for cxxopts already includes this patch since we need it for the xeus-cling kernel.

jarro2783:
Sorry I didn't see this. I don't see why this is necessary. See the comments on #69.

Also I was trying to avoid rebuilding the regex multiple times, which is why I put it outside the scope of the function that uses it.

jarro2783:
I managed to build `xeus-cling`, but didn't get any link errors. Is this only on a specific version of GCC?

SylvainCorlay:
Ok, I did some testing and the link error only occurs when I build `xeus-cling` with conda-build (which runs the cmake build command after setting some environment variables). The manual build of xeus-cling on my local machines does not trigger the erros.

jarro2783:
Hmm that's interesting. It took me a bit of fiddling around to get the build to work. Can you give me the minimal commands that will run the build that fails?

It also might help to give me the compiler command that is actually run, I might see something in there.

PR #75, opened 2017-10-09
avoid warning about non-virtual destructor

bebuch:
´Value´ and ´standard_value´ don't have a virtual destructor. When I understand the code right, this is intended, because they are used through ´shared_ptr´ and so its not required. Nevertheless, clang does warn about it, since at the point of the destructor call it can not check if the object has a final type. Adding the C++11 keyword ´final´ to ´standard_value´ avoids this warning.

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/75)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -694,7 +694,7 @@ namespace cxxopts
     };
 
     template <typename T>
-    class standard_value : public Value
+    class standard_value final : public Value
     {
       public:
       standard_value()


jarro2783:
Thanks. With which Clang and with what compiler flags does this happen?

bebuch:
It happens since the current development branch (clang 6) with `-Wall`. ; -)

```
$ clang++ --version
clang version 6.0.0 (http://llvm.org/git/clang.git c64cb39314a24eeab8f0f13f579b27c7a8e96662) (http://llvm.org/git/llvm.git c5599f675fb616a08e8f55b3eec79dc643a919c5)
Target: x86_64-unknown-linux-gnu
Thread model: posix
```

Thanks for merging and of course the library itself!

PR #79, opened 2017-10-31
Accidentally disables -Wnon-virtual-dtor for unrelated code

Alcaro:
The top of the file tests for \_\_GNUC\_\_ to disable -Wnon-virtual-dtor.

The bottom of the file tries to test for \_\_GNUC\_\_ to reset that warning to the default state. Unfortunately, it's typoed as \_\_GNU\_\_, which doesn't exist, so the warning is left disabled.

Simple typo, simple fix.

(surprised that gcc doesn't throw a warning that the warning stack isn't empty, but oh well)

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/79)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -1744,7 +1744,7 @@ Options::group_help(const std::string& group) const
 
 }
 
-#if defined(__GNU__)
+#if defined(__GNUC__)
 #pragma GCC diagnostic pop
 #endif
 


jarro2783:
Thanks for the pull request. I'm actually getting rid of this in the next release because I've fixed the underlying problem. But I'll merge this into master so that 1.x isn't broken.

PR #81, opened 2017-11-21
Compile warning fixes

jddecarlo:
My build settings are set to treat warnings as error, so I had to fix a couple of warnings in this library.  The changes are minor.

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/81)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -566,7 +566,7 @@ namespace cxxopts
           throw argument_incorrect_type(text);
         }
 
-        result = result * base + digit;
+        result = static_cast<US>(result * base + digit);
       }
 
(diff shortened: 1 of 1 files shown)


include/cxxopts.hpp
```diff
@@ -1091,7 +1091,7 @@ namespace cxxopts
 
       auto riter = m_results.find(iter->second);
 
-      return riter->second.count();
+      return static_cast<int>(riter->second.count());
```
jarro2783: Actually this would be better with the return type being `size_t` since it can never be negative.

jarro2783:
This has been superceded by #89 and 24162899c99206167d8361c0bbb8b534bd0fe072.

PR #87, opened 2017-12-03
`|` should be inside of `(…)`

objectx:
It's possible to use `(t|true|T|True)` as `truthy_pattern` but still failing `true` as a truthy under my environment :-(

```
Apple LLVM version 8.0.0 (clang-800.0.42.1)
Target: x86_64-apple-darwin15.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

\# clang-5.0.0 works as expected even if does not modify it.

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/87)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -446,7 +446,7 @@ namespace cxxopts
       std::basic_regex<char> truthy_pattern
         ("(t|T)(rue)?");
       std::basic_regex<char> falsy_pattern
-        ("(f|false|F|False)?");
+        ("((f|F)(alse)?)?");
     }
 
     namespace detail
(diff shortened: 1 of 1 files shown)


jarro2783:
This probably explains the failures I saw with clang in my travis build previously. It uses an older version of `libc++`, so I'm not sure if it got precedence wrong. I'll just test this a bit myself and I'll merge it if there are no problems with other environments. Travis passed on this PR so that's a good start.

Which version of `libc++` are you using?

jarro2783:
Also this probably applies to the false pattern too.

jarro2783:
I added clang in again and built this. See https://travis-ci.org/jarro2783/cxxopts/jobs/311008566. I think just do the same for `false` and it should be all good.

objectx:
`otool -L example` result is below:
```
cmake-build-debug/src/example:
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 120.1.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1226.10.1)
```

I've applied same fix to `false` pattern and update PR.

PR #89, opened 2017-12-14
fixed “possible loss of data” warning on Visual Studio

zethon:
`int` and `size_t` are different sizes on Windows, which generates a warning in Visual Studio. There's seems no reason NOT to return a `size_t` here.

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/89)


jarro2783:
Yeah that'll never be negative.

PR #94, opened 2018-01-15
Add default "default" value to boolean options

jgonzalezdr:
Add default "default" value of "false" to boolean options, therefore allowing to call `result["boolOpt"].as<bool>()` without throwing an exception.

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/94)


jgonzalezdr:
The build has some jobs failing because of Travis problems installing upgrades, please rebuild again.

PR #98, opened 2018-02-25
Added const to argv type for better interoperability.

jgonzalezdr:
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/98)

### DIFF
--- a/test/catch.hpp
+++ b/test/catch.hpp
@@ -10280,7 +10280,7 @@ namespace Catch {
 #ifndef __OBJC__
 
 // Standard C/C++ main entry point
-int main (int argc, const char * argv[]) {
+int main (int argc, char * argv[]) {
     return Catch::Session().run( argc, argv );
 }
 
--- a/test/catch.hpp
+++ b/test/catch.hpp
@@ -10280,7 +10280,7 @@ namespace Catch {
 #ifndef __OBJC__
 
 // Standard C/C++ main entry point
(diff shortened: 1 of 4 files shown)


jarro2783:
This doesn't pass the tests. `argv` actually needs to be changed.

jgonzalezdr:
@jarro2783: Compilation is now fixed. BTW, I'll also create a PR for the master branch.

test/catch.hpp
```diff
@@ -10280,7 +10280,7 @@ namespace Catch {
 #ifndef __OBJC__
 
 // Standard C/C++ main entry point
-int main (int argc, char * argv[]) {
+int main (int argc, const char * argv[]) {
```
jarro2783: This whole file is copied verbatim from the `catch` distribution and so is likely to be overwritten. You can probably leave this one out.

jgonzalezdr: You're right, I went a bit too happy with Find+Replace!

jarro2783:
:lgtm:

---

Reviewed 1 of 1 files at r1, 2 of 3 files at r2, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/jarro2783/cxxopts/98#-:-L76QYUB7dXGc2cxNfpJ:bnfp4nl)*

PR #99, opened 2018-02-27
Added const to argv type for better interoperability.

jgonzalezdr:
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/99)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -1079,7 +1079,7 @@ namespace cxxopts
     ParseResult(
       const std::unordered_map<std::string, std::shared_ptr<OptionDetails>>&,
       std::vector<std::string>,
-      int&, char**&);
+      int&, const char**&);
 
     size_t
     count(const std::string& o) const
@@ -1122,7 +1122,7 @@ namespace cxxopts
     get_option(std::shared_ptr<OptionDetails>);
 
     void
-    parse(int& argc, char**& argv);
+    parse(int& argc, const char**& argv);
 
     void
     add_to_option(const std::string& option, const std::string& arg);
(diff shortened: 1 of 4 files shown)


jarro2783:
The same comment as for #98, you don't want to modify `catch.hpp`.

jgonzalezdr:
Fixed.

SylvainCorlay:
@jarro2783 I just wanted to give a heads up that this was a backward incompatible change between `2.1.0` and `2.1.1`.

We were passing a `const char**` to `parse`, and met compilation errors when we updated to `2.1.1`.

(Turns out that 2.1.0 that was packaged for conda right when if got released and 2.1.1 much later, which is why we only encounter the issue today.)

SylvainCorlay:
Seems related to #123. For now, we are const-casting in xeus-cling to work around the issue.

jarro2783:
Yeah I reverted this for `2.1.1` because `2.1.0` was broken for more people. I'm planning on adding a separate `const` interface because there is no one size fits all.

SylvainCorlay:
Actually, const casting also makes it incompatible with `2.1.0` 

(`error: invalid conversion from ‘char**’ to ‘const char**`)

So these two versions are apparently mutually exclusive.

SylvainCorlay:
(basically I am assuming that 2.1.1 will remain the only version to not allow for `const char**`).

SylvainCorlay:
arg, these macros were only added after 2.1.1.

jarro2783:
I think it would be safer to assume for now that `2.1.0` is broken and that the constness will be fixed at some point.

SylvainCorlay:
Thanks. I now require `>=2.1.1,<3` and do the const cast.

PR #105, opened 2018-04-03
Allow unrecognised options.

fschoenm:
For our use of cxxopts we needed support for options that are not recognized by the parser. As this was still on the project's TODO list, I implemented a proposal for that function. It can be used like this:

    options
      .allow_unrecognised_options()
      .add_options()
        ("a,apple", "an apple", cxxopts::value<bool>(apple))
        ("b,bob", "Bob")

The ``Options`` and ``ParseResult`` objects remember this setting and skip unknown arguments instead of throwing an exception if set to ``true``.

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/105)

### DIFF
--- a/test/options.cpp
+++ b/test/options.cpp
@@ -483,9 +483,10 @@ TEST_CASE("Unrecognised options", "[options]") {
 
   Argv av({
     "unknown_options",
+    "--unknown",
     "--long",
     "-su",
-    "--unknown",
+    "--another_unknown",
   }); 
 
   char** argv = av.argv();
@@ -498,5 +499,7 @@ TEST_CASE("Unrecognised options", "[options]") {
   SECTION("After allowing unrecognised options") {
     options.allow_unrecognised_options();
     CHECK_NOTHROW(options.parse(argc, argv));
+    REQUIRE(argc == 3);
+    CHECK_THAT(argv[1], Catch::Equals("--unknown"));
   }
 }
--- a/test/options.cpp
+++ b/test/options.cpp
@@ -473,3 +473,30 @@ TEST_CASE("std::optional", "[optional]") {
   CHECK(*optional == "foo");
 }
 #endif
+
+TEST_CASE("Unrecognised options", "[options]") {
+  cxxopts::Options options("unknown_options", " - test unknown options");
+
+  options.add_options()
+    ("long", "a long option")
+    ("s,short", "a short option");
+
+  Argv av({
+    "unknown_options",
+    "--long",
+    "-su",
+    "--unknown",
+  }); 
+
+  char** argv = av.argv();
+  auto argc = av.argc();
+
+  SECTION("Default behaviour") {
+    CHECK_THROWS_AS(options.parse(argc, argv), cxxopts::option_not_exists_exception);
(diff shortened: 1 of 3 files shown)


jarro2783:
Thanks for the pull request. This looks about right. To clarify your intent here: when `-o` or `--option` are seen and those are not registered options, it just ignores them, leaving them in `argv` for the user to process later?

Can you write a test for this too?

test/options.cpp
```diff
@@ -473,3 +473,30 @@ TEST_CASE("std::optional", "[optional]") {
...
+  const char** argv = av.argv();
+  auto argc = av.argc();
+
+  SECTION("Default behaviour") {
+    CHECK_THROWS_AS(options.parse(argc, argv), cxxopts::option_not_exists_exception);
+  }
+
+  SECTION("After allowing unrecognised options") {
+    options.allow_unrecognised_options();
+    CHECK_NOTHROW(options.parse(argc, argv));
```
jarro2783: It might be good to check that `argc` is still 4.

fschoenm: Well, right now ``argc`` won't be 4 but will be changed just as before by changes (so I believe it changes to 1 in the test case). I wasn't sure about the semantics of modifying ``argc``'s value so I didn't touch that behavior. It doesn't seem to be documented anywhere.

Of course I can change it back to the original value in case the parsing detects any unspecified options. Is that the behavior you would recommend?

jarro2783: It makes sense for it to be 4 if it's skipping unrecognised options. Although I just realised that I'm not sure if the code that consumes arguments will work with this.

fschoenm:
Do you have an idea how to continue from here?

jarro2783:
Sorry I forgot about this one. I'll take a look at this tomorrow and get back to you.

jarro2783:
At the moment, the code that consumes arguments is basically going up to a `--`, and then moving everything in `argv` down from there. So I think if you're going to skip unrecognised arguments you'll have to enhance the logic there to copy unrecognised arguments.

For example, if we have the following where all the short ones are unrecognised:

    --long -a --something -b -- more arguments

Then the end result should be:

    -a -b more arguments

fschoenm:
So the parser should consume all arguments it recognizes and remove them from ``argv``? Is that how it works at the moment? Because when I tried it with a simple case, the ``argv`` array didn't seem to be changed even though the ``argc`` value was.

What is supposed to happen if short options are consolidated? For example if only options ``-a`` and ``-c`` are known and you get the following argument: ``-abcd``. Do you want to get ``-bd`` as remainder because that would probably complicate the change somewhat.

jarro2783:
It kind of pretends to remove them by copying the unrecognized arguments to
the front and setting argc.

I hadn't thought of grouped short options, that gets messy. We could always
say we don't support that, or only if they're all together.

How about for now just do long options. I'll accept that and I can work on
it a bit further.

fschoenm:
I added that behavior for the long options and adapted the test. If your other tests are comprehensive it shouldn't have broken anything … :-)

PR #112, opened 2018-04-12
Use Absolute Paths Instead of Relative Paths for CMake Configuration

arnavb:
This is much less likely to break code if the file is moved for some reason.

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/112)

### DIFF
--- a/test/add-subdirectory-test/CMakeLists.txt
+++ b/test/add-subdirectory-test/CMakeLists.txt
@@ -5,7 +5,7 @@ project(cxxopts-test)
 set(CMAKE_CXX_STANDARD   11)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
-add_subdirectory(${PROJECT_SOURCE_DIR} cxxopts EXCLUDE_FROM_ALL)
+add_subdirectory(../.. cxxopts EXCLUDE_FROM_ALL)
 
 add_executable(library-test "${PROJECT_SOURCE_DIR}/src/example.cpp")
 target_link_libraries(library-test cxxopts)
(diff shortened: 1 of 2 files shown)


arnavb:
@jarro2783 That seems.... unusual. I'm going to play around with this to see what's wrong.

arnavb:
Ok, I've managed to get rid of the infinite loop, but now I have another problem (sigh). Apparently, CMake doesn't seem to be recognizing the top level `CMAKE_SOURCE_DIR` (Which is the directory that contains root `CMakeLists.txt`). For the life of me, I can't figure out why this is the case. If I were to take a wild guess, then I would presume that it's because the tests themselves are their own separate project, which is then the top level directory. Otherwise, I can't say. For now, I'm going to close this...

jarro2783:
Thanks for trying. I'm not so good with CMake myself, and someone else rewrote most of the CMake code for me in another pull request, so I'm not sure myself why it's not working.

jarro2783:
Actually I just had another look at your change and realised that you were changing the test files. I think since they are their own tests that are actually running `cmake` on the test directory, the test directory itself is the top level root. So probably the only way to improve this would be to make them template files that are rendered with the project directory at test time.

niosHD:
I can confirm that the those two test directories are completely detached from the overall project. The are used in a separate CMake invocation when running the tests and therefore become the root directory themselves.

The reason these relative paths are there is that the tests reuse the main project as well as the example application. I personally do not really have any problem with the relative paths in this case given the CI server would fail immediately when they get stale. However, if you think it is a serious problem I can look into the issue.

Disclaimer: I am the guy who did the last CMake rewrite and wrote these tests. Also, @jarro2783 feel free to ping and assign me on any CMake related issues and questions. I am more than happy to help with them. ;)

PR #128, opened 2018-09-14
fixed version parsing in the CMake code

niosHD:
The CMake code that has been introduced in cde83be99b6308ccbaf38e2757cdb96386d0715d for parsing the version from the header seems to be broken. Note the curly brace in the regex and the set command. This commit fixes the problem for me.

On a related note, are you interested in having package manager specific files in your upstream repository? I just packaged your library with [conan](https://docs.conan.io/en/latest/) since I needed it for a project. If you want I can send a PR with the required `conanfile.py`. (see https://github.com/niosHD/cxxopts/blob/master/conanfile.py)

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/128)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,20 +18,21 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 cmake_minimum_required(VERSION 3.1)
-project(cxxopts)
 
-enable_testing()
-
-file(STRINGS "${PROJECT_SOURCE_DIR}/include/cxxopts.hpp" cxxopts_version_defines
-     REGEX "#define {CXXOPTS__VERSION_(MAJOR|MINOR|PATCH)")
+# parse the current version from the cxxopts header
+file(STRINGS "${CMAKE_CURRENT_SOURCE_DIR}/include/cxxopts.hpp" cxxopts_version_defines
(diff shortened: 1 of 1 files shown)


niosHD:
> It looks reasonable, but I just want to clarify what exactly is broken.

Sorry, I am not sure if this is a question but I clarify just in case. Like noted in the first message, there is an incorrect curly brace in the regex string (line 28 between `#define` and `CXXOPTS...`) and another one at the beginning of the subsequent set command (line 29).

~~~CMake
if(ver MATCHES "#define {CXXOPTS__VERSION_(MAJOR|MINOR|PATCH) +([^ ]+)$")
    set({CXXOPTS__VERSION_${CMAKE_MATCH_1} "${CMAKE_MATCH_2}" CACHE INTERNAL "")
~~~

The remaining changes are just a little bit of clean up. As well as a debug message (i.e., `message(STATUS "cxxopts version ${VERSION}")`) which should probably be removed.

jarro2783:
I worked out how to see the problem and verify that it is fixed.

jarro2783:
Sure, I'll try to remember that. As far as packaging goes, I don't mind too much. If someone is asking for it then it's probably going to be helpful to more people.

I did have another package manager configuration in there for a while and that package manager ended up being abandoned. As long as other people can maintain it because I won't have a clue about every packager manager that exists.

niosHD:
Right, I remember removing the biicode stuff when I did the last CMake refactoring. After some cleanup, I am going to send a PR for conan support then within the next few days.

PR #140, opened 2018-10-10
Use quotations for include

keanemind:
The readme is telling users to use brackets for the include when they should be using quotation marks, if I am not mistaken.

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/140)

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Additionally, anything after `--` will be parsed as a positional argument.
 
 ## Basics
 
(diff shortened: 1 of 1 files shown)


jarro2783:
I would generally be expecting this to be installed in a system location, not the root of your own project. In that case the brackets are correct.

I would also expect that if someone wants to put it in a local path that they would know the difference, so I don't really want to just change this because it seems like a somewhat arbitrary choice.

PR #141, opened 2018-10-10
Fixed result string and line size

mmanolescu:
`result` string needs to get appended so that the whole text is shown, not just the last part. Also line `size` needs to get reset.

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/141)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -1417,8 +1417,10 @@ namespace cxxopts
 
         if (*current == '\n')
         {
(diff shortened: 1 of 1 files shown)


jarro2783:
Can you show me a before and after of some text that is broken?

mmanolescu:
Hello and sorry for the late reply.
Here is before:
image
Here is after:
image

jarro2783:
I believe this has been fixed somewhere else.

PR #149, opened 2018-11-19
Fix invalid exception type in catch

jeanguyomarch:
g++ 8.2.0 refuses to compile the test suite because of polymorphic types
that are caught by value. They shall be caught by reference instead.

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/149)


jarro2783:
Thanks.

PR #152, opened 2018-12-19
fix min/max defined macros

shuax:
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/152)


jarro2783:
What is this fixing?

shuax:
> What is this fixing?
Avoid this problem
https://stackoverflow.com/questions/5004858/stdmin-gives-error

jarro2783:
I guess this isn't so bad.

Issue #95, opened 2018-01-25
Error when setting a long string as the default value


wykvictor:
I'm just curious about the following lines to control the length of the option's description:

https://github.com/jarro2783/cxxopts/blob/76717cb3dd23fe77ad5d8ba2b7479b3eecc719a1/include/cxxopts.hpp#L1861-L1864

I have a very long string as the default value, and error occurs then calling options.help():
libc++abi.dylib: terminating with uncaught exception of type std::length_error: basic_string
Abort trap: 6

So I change size_t{76} to size_t{120}. What's the risk of doing this?
Thanks!

jarro2783:
That's to fit the options into the width of a standard terminal. Can you show me some code that reproduces the problem?

j2bbayle:
I think this is the kind of bug @wykvictor was pointing out:

```cpp
#include <iostream>
#include "cxxopts.hpp"

int main(int argc, char * argv[]) {

    cxxopts::Options parser = cxxopts::Options("Reproducible Bug");
    parser.add_options()
    ("h,help", "show this help message and exit")
    ("my-option", "description with long wooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooord", cxxopts::value<int>());

    auto args = parser.parse(argc, argv);
    if (args.count("help"))
    {
        std::cout << parser.help() << std::endl;
        exit(0);
    }

    return 0;
}
```
 which compiles, but throws a runtime error when executed with `--help`.

j2bbayle:
The issue should be named "Error for descriptions with long words".

lazysegtree:
I can't reproduce it anymore. Its likely fixed via 
- https://github.com/jarro2783/cxxopts/pull/218
or
- https://github.com/jarro2783/cxxopts/pull/215

@j2bbayle @wykvictor Can you confirm if you don't see the issue in latest cxxopts and this issue can be closed?

j2bbayle:
Thanks for looking at this. 

This issue is more than 7 year old, and it’s been forever I haven’t used cxxopts. I won’t be able to validate this…

PR #167, opened 2019-02-22
Break words if longer than allowed length

j2bbayle:
If a word is longer than the allowed length in an option's description, the program crashes when formatting it with the following exception (c.f. issue #95):
```
libc++abi.dylib: terminating with uncaught exception of type std::length_error: basic_string
Abort trap: 6
```

The algorithm insert words until the maximum allowed length is reached; then we insert a new line `\n` character and start again from the previous space character (i.e. the word that could not fit on the previous line). We therefore keep a pointer to the first character of the current line and the previous space.

If a word is longer than the allowed length, it will break the previous line, but cannot fit on the next one. The algorithm will try to insert the first part (with exactly the maximum allowed length), but the previous space is still on the previous line.

Setting the last space character to the first character of the new line prevents that bug.

Solve #95.

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/167)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -1456,6 +1456,7 @@ namespace cxxopts
             stringAppend(result, "\n");
             stringAppend(result, start, ' ');
             startLine = lastSpace + 1;
+            lastSpace = startLine;
           }
           size = 0;
         }


h3ndrk:
Hi, what's the status of this PR? I would love to see this PR merged, if possible, since I have the same problem of setting long strings as default values.

Thanks in advance!

j2bbayle:
On my side it is working, I don't see why it shouldn't be merged…

jarro2783:
Sorry I've been a bit busy, I'll take a look at this in the morning.

jarro2783:
Sorry this PR dropped off my radar, but it appears that the original issue was already fixed in 943cdcfbbaa2.

If you're happy with that you can close it.

h3ndrk:
I did two tests: With vcpkg@2020.06 (which uses cxxopts@2.2.0) the problem exists. When using cxxopts@master nothing breaks. So I can confirm that the master seems to be fixed for my use case (I will patch my vcpkg s.t. it uses a more recent version of cxxopts).

Is it possible that you publish a new release of cxxopts containing the fix s.t. we can update vcpkg with the new release upstream?

<details>
<summary>vcpkg testing example</summary>

### main.cpp

```cpp
#include <cxxopts.hpp>
#include <iostream>
#include <string>

int main(int argc, char *argv[]) {
  cxxopts::Options options{"test", "Testing long strings"};

  options.add_option("", "h", "help", "Print help", cxxopts::value<bool>(), "");

  std::string test{"abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefg"
                   "hijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz"};
  options.add_option("", "", "test", "Test flag with long default value",
                     cxxopts::value<decltype(test)>(test)->default_value(test),
                     "test");

  const auto result = options.parse(argc, argv);

  if (result.count("help") != 0) {
    std::cout << options.help() << std::endl;
    return 0;
  }

  return 0;
}
```

### CMakeLists.txt

```cmake
cmake_minimum_required(VERSION 3.10)

project(dataset LANGUAGES CXX)

find_package(cxxopts CONFIG REQUIRED)

add_executable(test
  "${CMAKE_CURRENT_SOURCE_DIR}/main.cpp"
)

target_link_libraries(test PRIVATE
  cxxopts::cxxopts
)
```

### setup.sh

Execute this script to install vcpkg, cxxopts and compile a simple program. You can trigger the bug with running `build/test --help`.

```bash
#!/bin/bash

git clone -b 2020.06 git@github.com:microsoft/vcpkg.git
vcpkg/bootstrap-vcpkg.sh -disableMetrics
vcpkg/vcpkg install cxxopts

mkdir build
cd build
cmake -DCMAKE_TOOLCHAIN_FILE=../vcpkg/scripts/buildsystems/vcpkg.cmake ..
make
```

</details>

j2bbayle:
Closing, then.

PR #168, opened 2019-03-07
Throw option_required_exception if option value is not present

sbogomolov:
Currently an std::domain_error is thrown in as(), but this does not show the missing option name. Proposed PR throws in ParseResult.operator[] when trying to get an option without a value.

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/168)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -1044,6 +1044,12 @@ namespace cxxopts
       return m_count;
     }
 
+    bool
+    has_value() const
+    {
+      return m_value != nullptr;
+    }
+
     template <typename T>
     const T&
     as() const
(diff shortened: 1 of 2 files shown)


jarro2783:
Looks reasonable. The tests will need fixing though.

sbogomolov:
Sure, my bad I have not checked the tests. Also, it might be worth a consideration to remove the check in as() method. Unless there is a way to obtain OptionValue other than with ParseResult.operator[], that check will never trigger. Let me know if you agree, I'll update the PR.

jarro2783:
I believe this has already been changed to something more suitable.

PR #172, opened 2019-04-15
Remove `u8` prefix in string literals.

le-migou:
String literals with the prefix `u8` now have the type `const char8_t[N]` in
C++20. As a consequence the library does not compile in c++2a mode with GCC
(you can't concatenate `std::basic_string<char>` and `const char8_t[]` as
`char` and `char8_t` are different types).

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/172)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -354,7 +354,7 @@ namespace cxxopts
   {
     public:
     option_exists_error(const std::string& option)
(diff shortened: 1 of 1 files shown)


jarro2783:
Seems reasonable. Unicode is a mess in C++ anyway and I'm not sure if there is any practical difference here.

PR #177, opened 2019-05-27
Parse 0 and 1 into booleans

j2bbayle:
Regular expressions to parse boolean options currently do not accept "0" nor "1". In my opinion, these values should respectively be parsed as "false" and "true", so that such commands `./myprogram --option=0` are valid.

Similarly, empty values are accepted and parsed as "false", which is error-prone. No error is raised when an argument is missing, eg. with `./myprogram --option` or `./myprogram --option=`. This is equivalent to setting "false" as the implicit default value, so that the behaviour can be recovered if necessary. I propose to remove this behaviour.

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/177)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -466,9 +466,9 @@ namespace cxxopts
       std::basic_regex<char> integer_pattern
         ("(-)?(0x)?([0-9a-zA-Z]+)|((0x)?0)");
       std::basic_regex<char> truthy_pattern
-        ("(t|T)(rue)?");
+        ("(t|T)(rue)?|1");
       std::basic_regex<char> falsy_pattern
-        ("((f|F)(alse)?)?");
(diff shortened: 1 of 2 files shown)


include/cxxopts.hpp
```diff
@@ -1456,6 +1456,7 @@ namespace cxxopts
             stringAppend(result, "\n");
             stringAppend(result, start, ' ');
             startLine = lastSpace + 1;
+            lastSpace = startLine;
```
jarro2783: what's this change for?

j2bbayle: That's a commit from #167. My bad. Fixed now...

jarro2783:
Can you add to the tests?

j2bbayle:
Sure, that should be good now.

PR #178, opened 2019-05-29
Add a method to remove the implicit value of an option

j2bbayle:
An option with an implicit value can be used without setting a value, such as `my_program --myoption`. However, if the user wishes to set a custom value, only the syntax with an equal sign will be recognized (`--myoption=myvalue`). The usual syntax using a space character to separate the option and the value (`--myoption myvalue`) is invalid, since it is not possible to disambiguate it from a positional argument.

This is not a problem in general, because the user has the choice to set an implicit value, or leave the option without an implicit value. In this last case, the syntax `my_program --myoption` is not valid, and `my_program --myoption myvalue` is correct.

In the case of boolean options, `true` is automatically set as the implicit value and there's currently no way to prevent that. This might be an issue if one wants to support the syntax `--myoption myvalue`.

I propose to add a method `no_implicit_value()`, so that the user can disable the automatic implicit value, which is set for boolean options. This is how it might be used:

```cpp
cxxopts::Options parser = cxxopts::Options("my_program");
parser.add_options()
    ("h,help", "show this help message and exit")
    ("with-implicit", "boolean with auto implicit value", cxxopts::value<bool>())
    ("no-implicit", "boolean without implicit value", cxxopts::value<bool>()->no_implicit_value());
```

so that we the following command throws an error:
```shell
my_program --no-implicit
```

but the following command is valid, and `with-implicit` is parse to `true`, `no-implicit` parsed to `false`:
```shell
my_program --with-implicit --no-implicit false
```

Unit tests are added to check this behaviour.

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/178)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -309,6 +309,9 @@ namespace cxxopts
     virtual std::shared_ptr<Value>
     implicit_value(const std::string& value) = 0;
 
+    virtual std::shared_ptr<Value>
+    no_implicit_value() = 0;
+
     virtual bool
     is_boolean() const = 0;
   };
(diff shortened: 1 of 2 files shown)


jarro2783:
Looks good to me.

PR #182, opened 2019-06-17
Fix parsing of std::vector and add test, example and documentation

langchr86:
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/182)


jarro2783:
The intent for vector was that you could write an option multiple times and keep all the values. But this seems like reasonable behaviour as well.

Issue #185, opened 2019-06-20
segfault when using undefined options


Remi-Laot-CreiZyz:
The following snippet results in a `SEGFAULT` when I do not input a `-i <...>` option running my program.

```
  try
  {
    cxxopts::Options options("xxx", "descr.");
    options.add_options("target")
      ("i,input",  "input (JSON format)", cxxopts::value<std::string>() )
      ("o,output", "output (CSV format)", cxxopts::value<std::string>() )
    ;
    auto namedOptions = options.parse(argc, argv);
    std::string inFile = namedOptions["input"].as<std::string>(); // << SEG FAULT
    
    // ...
  
  }
  catch(const cxxopts::OptionException& e)
  {
    std::cerr << "error parsing options : " << e.what() << std::endl;
    exit(EXIT_FAILURE);
  }
```

Am I doing something wrong or is this a bug ?

Thank you for your time.

eamanu:
Hello @Remi-Laot-CreiZyz,

I can confirm the behavior :/

eamanu:
if you add:

```
if (namedOptions.count("input")){
     std::string inFile = namedOptions["input"].as<std::string>();
}
```

you will not have the segfaults :-)

jarro2783:
If you try to get the value of an option that hasn't been defined it will crash. This is expected, although an exception would probably be better than a segfault.

jarro2783:
Actually looking at the code I didn't think it would do that. I'll try to reproduce this myself.

jarro2783:
Mine doesn't segfault:
```
src/example
terminate called after throwing an instance of 'std::domain_error'
  what():  No value
```

Can you provide a full snippet that reproduces it?

eamanu:
if you compile:

```
#include "cxxopts.hpp"
#include <iostream>

int main(int argc, char *argv[]){
  try
  {
    cxxopts::Options options("xxx", "descr.");
    options.add_options("target")
      ("i,input",  "input (JSON format)", cxxopts::value<std::string>() )
      ("o,output", "output (CSV format)", cxxopts::value<std::string>() )
    ;
    auto namedOptions = options.parse(argc, argv);
    std::string inFile = namedOptions["input"].as<std::string>(); // << SEG FAULT
  }
  catch(const cxxopts::OptionException& e)
  {
    std::cerr << "error parsing options : " << e.what() << std::endl;
    exit(EXIT_FAILURE);
  }

}
```

this produce a segfaults

[...]

PR #186, opened 2019-06-24
Create a exception to thow when Options not exists

eamanu:
When the user try to get a option that not exist, cxxopt library
throw a std::domain_error, that make a segfaults. See #185.

On this PR I create a new Exception class from OptionParseException
that is throw when this situation occur. This avoid the SegFaults
error.

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/186)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -394,6 +394,15 @@ namespace cxxopts
     }
   };
 
+  class option_value_not_exists_exception : public OptionParseException
+  {
+    public:
+    option_value_not_exists_exception()
+    : OptionParseException("Option value does not exist")
+    {
+    }
+  };
+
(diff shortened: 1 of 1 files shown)


jarro2783:
I don't think it's a `ParseException`, so it probably shouldn't derive from that. I would like to know why it ends up segfaulting. Presumably if the `std::domain_error` is caught then it won't segfault.

I think this one makes sense to be a `std::domain_error`, in the same way that you'll get a dynamic cast exception if you cast a value that exists to the wrong type. I don't think that this error is using cxxopts incorrectly in the same way that all the other exceptions indicate parse errors, at that point you're just using a value stored in a container incorrectly.

jarro2783:
Closing because this is currently the correct behaviour.

PR #189, opened 2019-06-26
add_options variant with initializer list

gogo2:
This adds variant of Options::add_options with initializer list rather than (in my opinion) strange looking(taking into account syntax features of modern c++) solution with OptionAdder and opeator()

To take advantage of Option struct I also added add_option(string, Option); variant.

Example:

```
options.add_options("", { 
{"a, address", "server address", cxxopts::value<std::string>()->default_value("127.0.0.1")},
{"p, port", "server port", cxxopts::value<std::string>()->default_value("7110")}
});


options.add_option("", {"h, help",    "help"});

cxxopts::Option option_d{"d,data", "Data"};
options.add_option("", option_d);
```

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/189)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -1220,6 +1220,14 @@ namespace cxxopts
 
     std::vector<KeyValue> m_sequential;
   };
+  
+  struct Option 
+  {
+    std::string opts;
+    std::string desc;
+    std::shared_ptr<const Value> value = ::cxxopts::value<bool>();
+    std::string arg_help = "";
+  };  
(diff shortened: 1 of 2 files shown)


jarro2783:
This looks like a good change. Can you add some tests?

gogo2:
Tests deivered with fix for c++11 compatibility(initial version worked only in 14)

jarro2783:
Looks great, thanks for your contribution.

PR #190, opened 2019-07-14
Allow disabling exceptions and use std::abort instead

charlydelta:
In a project I am required to compile with `-fno-exceptions`, preventing the use of this library. I guarded all throw statements with a def `CXXOPTS_NO_EXCEPTIONS` in the same style as  `CXXOPTS_NO_RTTI` and replaced it with `std::abort()` if exceptions are to be disabled

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/190)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -495,14 +495,22 @@ namespace cxxopts
           {
             if (u > static_cast<U>((std::numeric_limits<T>::min)()))
             {
+#ifndef CXXOPTS_NO_EXCEPTIONS
               throw argument_incorrect_type(text);
+#else
+              std::abort();
+#endif
             }
           }
           else
           {
             if (u > static_cast<U>((std::numeric_limits<T>::max)()))
             {
+#ifndef CXXOPTS_NO_EXCEPTIONS
               throw argument_incorrect_type(text);
+#else
+              std::abort();
+#endif
             }
           }
         }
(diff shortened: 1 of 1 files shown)


jarro2783:
Whilst I agree that exceptions are not always appropriate for everyone, I'm not so sure that this is necessarily the best change, although it might do for now.

If a user passes an invalid option then this will abort the whole program with no error message, is that desirable in your case?

charlydelta:
Thank you very much for your feedback. 

> Whilst I agree that exceptions are not always appropriate for everyone, I'm not so sure that this is necessarily the best change, although it might do for now.

I agree, there might be better changes, but this means individual solutions for each substitution. For example, the exception in `ParseResult::operator[]` cannot be circumvented. Here, a method `ParseResult::has` or so would be necessary, so it can be checked in advance whether the call to `operator[]` will be save. (Let me know if I should add such a method for this case, and if so, what name. Also, if you have solutions for other occurrences, let me know.)

> If a user passes an invalid option then this will abort the whole program with no error message, is that desirable in your case?

Of course error messages would be better, I totally agree. I just thought to make a minimal PR as foundation for further discussions. I would solve this by printing `text` to `stderr` before abort. This would also solve another problem I encountered after opening this PR, namely that it can't be compiled because of `-Werror` and unused parameters `text` if the exception was replaced.

jarro2783:
Ok I think we'll merge this as is because it will do for now and it can be improved later.

charlydelta:
Like I already said, I'm afraid this won't compile because of `-Werror` and unused parameters.

Printing the error message to `std::cerr` would solve this. Need a PR for this? I can do that today evening.

jarro2783:
Oh right I didn't notice that. Another PR would be great.

charlydelta:
Will do when I come home

PR #200, opened 2019-08-20
Get default value of boolean from get()

maxint:
It's more handy for boolean parameters with default values.
```c++
struct Options {
  bool foo = true;
  double bar = 1;
};
Options options = {};
cxxopts::Options parser("foobar");
parser.add_options()
  ("f,foo", "foo", cxxopts::value<bool>(options.foo))
  ("b,bar", "bar", cxxopts::value<double>(options.bar))
  ;
```

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/200)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -949,7 +949,7 @@ namespace cxxopts
       set_default_and_implicit()
       {
         m_default = true;
-        m_default_value = "false";
+        m_default_value = get() ? "true" : "false";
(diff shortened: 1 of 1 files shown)


jarro2783:
I'm not sure what problem this is trying to solve. Can you make a test case that illustrates the behaviour that this adds?

jarro2783:
Closing as no activity. Please reopen and address comments if needed.

PR #203, opened 2019-08-23
Disable tests if cxxopts is used via add_subdirectory

icpz:
Disable tests/examples/install targets when cxxopts is used as bundled library,
e.g. via `add_subdirectory`.

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/203)


jarro2783:
I think this was addressed with #270. Can you open a new PR if there is still a problem?

PR #204, opened 2019-09-09
allow empty string to be valid positional arguments

wangbj:
*cxxopts* doesn’t accept empty string from positional argument, but silently disregard it instead, this should’ve been the ideal solution, i.e. when running `ssh-keygen -t rsa -N “”`. Passing `\”\”` wouldn’t help either. This PR tries to allow passing empty string for positional arguments.


wangbj:
sorry for update late, added a test case and it passes iff the fix is in.

Issue #217, opened 2020-01-04
Assertion failure on Options.help() "string iterator range transposed"


Araeos:
OS: Windows 10
Compiler: Visual Studio 2019
Version: Master (23f56e2)
When calling the member method help() of the Options instance, an assertion fails during debug. The issue is caused when `cxxopts::stringAppend(s, begin, end)` is called where `end < begin`. I will try a fix in a pull request.

PR #218, opened 2020-01-04
Fix assertion failure (issue #217)

Araeos:
This fix does work in my case, but someone should verify it does not break some other expected behaviour.
See Issue #217 for more information.

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/218)


jarro2783:
This looks reasonable. I'll compile it and check a few cases to make sure it looks like it's doing the right thing.

PR #221, opened 2020-02-14
sun hack

despair86:
plain char is signed on Oracle's IX86-based hardware (and legacy x86 Sun workstations) just like every other OS that runs on the Intel Pentium Pro, but compilers on this platform _**do not**_ treat plain `char` as a type distinct from `signed` /`unsigned char` such that `int8_t` and `char` resolve to the same type, triggering an error in compilation

This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/221)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -39,6 +39,10 @@ THE SOFTWARE.
 #include <unordered_set>
 #include <vector>
 
+#ifdef __sun
+#define int8_t signed char
+#endif
+
 #ifdef __cpp_lib_optional
 #include <optional>
 #define CXXOPTS_HAS_OPTIONAL


eyalroz:
1. Why does this need to be part of `cxxopt`?
2. It seems like a bad idea to preprocessor-define an identifier which is a standard C type.

despair86:
yeah i left this hanging a while back and forgot about it, sorry

despair86:
for reference this is what Netscape said about this some 20 years back:
```c
/*
** There are two scenarios that require us to define PRInt8 as type 'char'.
** (1)
** Some cfront-based C++ compilers do not like 'signed char' and
** issue the warning message:
**     warning: "signed" not implemented (ignored)
** For these compilers, we have to define PRInt8 as plain 'char'.
** Make sure that plain 'char' is indeed signed under these compilers.
** (2)
** Mozilla C++ code expects the PRInt{N} and int{N}_t types to match (see bug
** 634793). If a platform defines int8_t as 'char', but NSPR defines it as
** 'signed char', it results in a type mismatch.
** On such platforms we define PRInt8 as 'char' to avoid the mismatch.
*/
```

PR #222, opened 2020-02-17
Support parsing value from ENV variable

deric:
Allows parsing ENV variables (at least on Linux/Unix).

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/222)

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,14 @@ completely replaced by calling `options.custom_help`. Note that you might
 also want to override the positional help by calling `options.positional_help`.
 
 
+## Value from ENV variable
+
+When parameter is not set, value will be fetched from an environment variable (if such variable is defined).
+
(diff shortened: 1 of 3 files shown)


jarro2783:
I'm not too keen on adding this right now. I want this library to stay lightweight and not have too many extra things like file parsing and environment variables.

I am considering writing a wrapper utility that might be a better place to put this sort of logic.

deric:
Sure, I understand. Though ENV variables are included in `std::` and doesn't add any extra dependency.

eyalroz:
@deric : Shouldn't you check, in `env()`, that `m_default` is off? (And perhaps `m_implicit` as well)?

jarro2783:
I don't want to implement this feature in this way so closing this one.

PR #232, opened 2020-05-20
Fix compile issue due to utf-8 symbols

Maximus5:
Visual Studio throws warnings because of LQUOTE and RQUOTE values:

    The file contains a character that cannot be represented in the current code page (936). Save the file in Unicode format to prevent data loss

Perhaps most proper solution is to add BOM mark.

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/232)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -1,4 +1,4 @@
-/*
+﻿/*
 
 Copyright (c) 2014, 2015, 2016, 2017 Jarryd Beck
 


neheb:
Weird. Which version is this?

jarro2783:
I don't think it should be necessary to add a BOM to a UTF-8 file. Is your Visual Studio correctly setup to automatically use UTF-8?

PR #234, opened 2020-06-02
[clang-tidy] use nodiscard

neheb:
Found with modernize-use-nodiscard

This should be safe as the standard says unknown attributes are to be
ignored.

Signed-off-by: Rosen Penev <rosenp@gmail.com>

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/234)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -45,6 +45,12 @@ THE SOFTWARE.
 #define CXXOPTS_HAS_OPTIONAL
 #endif
 
+#if __cplusplus >= 201603L
+#define CXXOPTS_NODISCARD [[nodiscard]]
+#else
+#define CXXOPTS_NODISCARD
+#endif
+
 #ifndef CXXOPTS_VECTOR_DELIMITER
 #define CXXOPTS_VECTOR_DELIMITER ','
 #endif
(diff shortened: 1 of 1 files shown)


neheb:
replaced with warn_unused_result since -Werror is being passed.

eyalroz:
@neheb : Your changes are non-standard. You probably want `[[nodiscard]]`, or rather:
```
#if __cplusplus >= 201712L
#define CXXOPTS_NODISCARD [[nodiscard]]
#else
#define CXXOPTS_NODISCARD 
#endif
```
and then use `CXXOPTS_NODISCARD`.

neheb:
Done

eyalroz:
@neheb : Nice :-) ... but - I had a typo. [It's actually 201703L](https://stackoverflow.com/a/38456243/1593077), not 201712L. Sorry...

neheb:
Fixed.

eyalroz:
Ok, but now - why have `[[nodiscard]]` on return values such as `bool` or `size_t`? I can understand having it for `shared_ptr`'s and such.

neheb:
It was an automated change.

neheb:
According to https://en.cppreference.com/w/cpp/feature_test , the proper value is 201603L . Changed accordingly.

PR #235, opened 2020-06-02
Improve help of short-only options (remove dangling comma)

akosthekiss:
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/235)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -1468,7 +1468,11 @@ namespace cxxopts
 
       if (!s.empty())
       {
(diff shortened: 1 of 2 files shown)


akosthekiss:
Adding a short-only option is supported, like this:
```cpp
("x", "A short-only option", cxxopts::value<std::string>())
```

However, current help of such options looks like the excerpt below (note the dangling comma):
```
  -x, arg                       A short-only option
```

After the change, the help looks like this:
```
  -x arg                        A short-only option
```

PR #246, opened 2020-07-21
Fixes #245: Mention the option name when throwing on "no value"

### LINKED ISSUE #245: When throwing due to a missing option value, the option name should be mentioned
When cxxopt's API is used to try and obtain an option value after parsing, but the option doesn't have a value, we currently get something like:
```
terminate called after throwing an instance of 'std::domain_error'
  what():  No value
```
The what-text is insufficient: it's too general and vague. The exception thrown should indicate which value was missing, e.g.
```
terminate called after throwing an instance of 'std::domain_error'
  what():  Command-line option 'foo' has no value
```

eyalroz:
* Added a new exception type: `option_has_no_value_exception`; throwing it when an option has no value we can cast with `as()`, instead of an `std::domain_error`.
* The `OptionValue` type now holds a pointer to the long option name (in its corresponding key within ParseResults's `m_results` field.
* Adjusted tests accordingly.

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/246)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -443,6 +443,18 @@ namespace cxxopts
     }
   };
 
+  class option_has_no_value_exception : public OptionException
+  {
+    public:
+    explicit option_has_no_value_exception(const std::string& option)
+    : OptionException(
+        option.empty() ?
+        ("Option " + LQUOTE + option + RQUOTE + " has no value") :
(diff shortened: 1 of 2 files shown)


jarro2783:
This seems reasonable. I'll take a look at it more closely a bit later.

PR #247, opened 2020-07-22
README.md elaboratin regarding repeated use of same option

eyalroz:
README will now explain how vectors of values allow repeated use of the same option on the command-line.

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/247)


jarro2783:
Thanks. I didn't realise that this wasn't clear enough.

PR #250, opened 2020-08-06
Fix argv const correctness in cxxopts::parse

kjetand:
When user code is working on a 'const char**' they have to do
a 'const_cast' in order to pass argv into the cxxopts::parse
function.

The rationale behind this commit is to accept both 'char**'
and 'const char**', but internally use 'const char**' due to
const correctness.

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/250)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ options. The project adheres to semantic versioning.
 * Fix duplicate default options when there is a short and long option.
 * Add `CXXOPTS_NO_EXCEPTIONS` to disable exceptions.
 * Fix char parsing for space and check for length.
+* Change argument type in `Options::parse` from `char**` to `const char**`.
 
 ## 2.2
 
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -1211,7 +1211,7 @@ namespace cxxopts
       >,
       std::vector<std::string>,
       bool allow_unrecognised,
-      int&, char**&);
+      int&, const char**&);
 
     size_t
     count(const std::string& o) const
@@ -1251,7 +1251,7 @@ namespace cxxopts
     private:
 
     void
-    parse(int& argc, char**& argv);
+    parse(int& argc, const char**& argv);
 
     void
     add_to_option(const std::string& option, const std::string& arg);
@@ -1274,7 +1274,7 @@ namespace cxxopts
     checked_parse_arg
     (
       int argc,
-      char* argv[],
+      const char* argv[],
       int& current,
       const std::shared_ptr<OptionDetails>& value,
       const std::string& name
@@ -1361,7 +1361,7 @@ namespace cxxopts
     }
 
     ParseResult
-    parse(int& argc, char**& argv);
+    parse(int& argc, const char**& argv);
 
     OptionAdder
     add_options(std::string group = "");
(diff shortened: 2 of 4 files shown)


include/cxxopts.hpp
```diff
@@ -1844,6 +1847,13 @@ Options::parse(int& argc, char**& argv)
 inline
 void
 ParseResult::parse(int& argc, char**& argv)
+{
+  parse(argc, const_cast<const char**&>(argv));
```
jarro2783, line 1851: I don't think you need this. You only need const cast to take away constness. Adding constness should never be a problem.

test/options.cpp
```diff
@@ -69,7 +69,7 @@ TEST_CASE("Basic options", "[options]")
     "--space",
   });
 
-  char** actual_argv = argv.argv();
+  const char** actual_argv = argv.argv();
```
jarro2783, line 72: might as well use `auto` here too

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ options. The project adheres to semantic versioning.
 * Fix duplicate default options when there is a short and long option.
 * Add `CXXOPTS_NO_EXCEPTIONS` to disable exceptions.
 * Fix char parsing for space and check for length.
+* Fix argv const correctness in cxxopts::parse
```
jarro2783, line 15: Can you make this comment clearer to describe exactly what constness is being improved?

include/cxxopts.hpp
kjetand: I'm not sure if I understand, could you elaborate? - Passing a 'char**' to a method that takes 'const char**' doesn't compile, hence the const_cast. E.g. reinterpret_cast will not work because "reinterpret_cast from type 'char**' to type 'const char**' casts away qualifiers".

kjetand:
> I just want to make sure that you understand what constness is being added here.
> 
> This ensures that the parser doesn't change the contents of the strings in `argv`, but it will still change the pointers around in the argv array of pointers. So this is probably correct, but won't do anything to address the concerns everyone has had about changing the argv array.

Thanks for the review :)

Yes, I understand the constness stuff, but the explaination could be better in the documentation (I will improve it). The concrete problem I wanted so solve was to unit test my application using hard coded strings. E.g. if I write `std::array argv{"myapp", "-h"};` the string literals are considered 'const char*', and this forced me to const_cast, which is ugly from a user perspective.

include/cxxopts.hpp
jarro2783: It compiled fine when I took away the `const_cast`. You can always pass non-const to const, it's the other way around that you can't do.

jarro2783:
> The concrete problem I wanted so solve was to unit test my application using hard coded strings. E.g. if I write `std::array argv{"myapp", "-h"};` the string literals are considered 'const char*', and this forced me to const_cast, which is ugly from a user perspective.

Yes this is not ideal. If it can be made to work with const char** then that's a start.

I am planning on rewriting some stuff so that it doesn't modify argv at all, and then it can be `char const * const * &`, but for now, this is a step in the right direction.

include/cxxopts.hpp
jarro2783: In fact you can take the `char**&` version away and leave the `const char**&` version.

kjetand: Fixed in 216f2ca

test/options.cpp
kjetand: Fixed in  216f2ca

CHANGELOG.md
kjetand: Fixed in  216f2ca

kjetand:
I've squashed and force pushed to my branch. The commit should be ready for merge if you accept (216f2ca).

PR #256, opened 2020-09-29
Fix empty option name in `OptionException`

### LINKED ISSUE #255: Empty option name on exception throwing
The following code:

   options.add_options()
        ("m,model", "Model file", cxxopts::value<std::string>())
        ("i,input", "Input video filename", cxxopts::value<std::string>())
        ("o,output", "Result video filename", cxxopts::value<std::string>())
        ("h,help", "Print usage")
        

    auto result = options.parse(argc, argv)

    a.model = result["model"].as<std::string>()

throwing exception ( as expected ) with empty option name:

` Option '' has no value`

spmn:
This should fix #255

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/256)


jarro2783:
This looks reasonable. I have a refactor coming shortly though, and I think it would be better rebased on that. You can look at the `parse_refactor` branch to see the changes. The branch is pretty stable now, so it should be safe to use.

jarro2783:
I've merged those changes now, you'll have to rebase this one.

jarro2783:
Sorry I forgot about this one. Can you rebase again?

jarro2783:
Actually never mind, I can fix this. I'll merge it shortly.

PR #258, opened 2020-10-01
Fix passing a const array to parse

jarro2783:
Fixes #257. The input array is not modified, so we can declare this as
`char const* const*`.

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/258)


jarro2783:
@sheerluck is this a better fix than #257?

sheerluck:
yes, this is better :)

Issue #263, opened 2020-10-19
potential null pointer dereference error with gcc-10. (gcc bug ?)


bsergean:
```
/xxx/build/_deps/cxxopts-src/include/cxxopts.hpp: In member function ‘size_t cxxopts::ParseResult::count(const string&) const’:
/xxx/build/_deps/cxxopts-src/include/cxxopts.hpp:1027:14: warning: potential null pointer dereference [-Wnull-dereference]
 1027 |       return m_count;
      |              ^~~~~~~
```

The code is:

```
    size_t
    count() const
    {
      return m_count;
    }

...
    // m_count being a member variable
    size_t m_count = 0;
```

I don't really get it, but reporting just in case. This is one of the few places where an integer is initialized like that, maybe initializing it in a constructor would get rid of the problem.

jarro2783:
This sounds like a bug in GCC, the only potential null pointer there is `this`, but then you would have that problem in _every_ member function. I would hope that GCC is not trying to solve that problem.

bsergean:
Yes that sounds like a bug in gcc. The reason this is getting in my way is that the examples are built by default and so when I include your project through cmake magic things break.

option(CXXOPTS_BUILD_EXAMPLES "Set to ON to build examples" ON)
option(CXXOPTS_BUILD_TESTS "Set to ON to build tests" ON)

If we set those to OFF, or just the example this would fix it.

ps: I don't think I'm alone in using cmake that way. I use it like that.

include(FetchContent)
FetchContent_Declare(cxxopts
    GIT_REPOSITORY "https://github.com/jarro2783/cxxopts"
    GIT_TAG "v2.1.2")

FetchContent_MakeAvailable(cxxopts)

And finally

target_link_libraries(${EXAMPLE_PROJECT_NAME} ${PROJECT_LIBRARIES} cxxopts ${PROJECT_NAME})

PR #273, opened 2021-01-08
Ignore gcc-10 warning

bsergean:
This warning trigger build errors, those pragma are a bit ugly but they prevent the build error.

```
In file included from /home/benjamin.sergeant/src/ALHttp-Cpp/build/_deps/cxxopts-src/src/example.cpp:27:
/home/benjamin.sergeant/src/ALHttp-Cpp/build/_deps/cxxopts-src/include/cxxopts.hpp: In member function ‘size_t cxxopts::ParseResult::count(const string&) const’:
/home/benjamin.sergeant/src/ALHttp-Cpp/build/_deps/cxxopts-src/include/cxxopts.hpp:1027:14: error: potential null pointer dereference [-Werror=null-dereference]
 1027 |       return m_count;
      |              ^~~~~~~
```

This should fix https://github.com/jarro2783/cxxopts/issues/263

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/273)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -1133,12 +1133,21 @@ namespace cxxopts
       m_value->parse();
     }
 
+#if defined(__GNUC__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Werror=null-dereference"
+#endif
+    
     CXXOPTS_NODISCARD
     size_t
     count() const noexcept
     {
       return m_count;
     }
+    
+#if defined(__GNUC__)
+#pragma GCC diagnostic pop
+#endif
 
     // TODO: maybe default options should count towards the number of arguments
     CXXOPTS_NODISCARD


jarro2783:
This looks like a bug. If this is a null pointer dereference, then so is _any_ member function.

I'm happy with the fix for now, but I will revisit if it is fixed. 

Do you have a bug reference for GCC?

jarro2783:
@bsergean which version of GCC was this for. It looks like for later versions of GCC this warning doesn't even exist.

bsergean:
$ gcc --version
gcc (Ubuntu 10.1.0-2ubuntu1~18.04) 10.1.0
Copyright (C) 2020 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

jarro2783:
I have this 10.2.0 and it fails with the same error as #276, so I'll have to try disable it globally.

jarro2783:
I don't see the error on the cxxopts build, do you have a specific flag turned on that produces this?

bsergean:
I think we see this when we make a release build

jarro2783:
Maybe we can add another check around it for the compiler version.

bsergean:
https://gcc.gnu.org/onlinedocs/cpp/Common-Predefined-Macros.html

bsergean:
btw the #pragma will disappear link appears to be a '-fpermissive' option will disappear which is different.

PR #274, opened 2021-01-17
Fixing cmake developer warning

jpr89:
Here is the warning currently being produced:

CMake Warning (dev) at C:/Program Files/CMake/share/cmake-3.19/Modules/GNUInstallDirs.cmake:223 (message):
Unable to determine default CMAKE_INSTALL_LIBDIR directory because no
target architecture is known. Please enable at least one language before
including GNUInstallDirs.
Call Stack (most recent call first):

I noted how I fixed the error. This is caused by GNUInstallDirs automatically executing code just by including it.

I also added -Werror=dev to the CI to ensure this never happens again.

I didn't catch this error before because I was consuming this project. But if you try to build it standalone you'll get this issue.

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/274)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,7 +61,7 @@ matrix:
             - g++-5
           sources: *sources
 script: >
-    cmake -DCXXOPTS_BUILD_TESTS=ON -DCMAKE_CXX_COMPILER=$COMPILER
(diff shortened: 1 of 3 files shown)


jarro2783:
Thanks. I noticed it too but wasn't sure how to fix it.

PR #283, opened 2021-04-21
-Wsuggest-override is not supported by gcc before 5.0

RonxBulld:
The -Wsuggest-override option is not supported in gcc prior to version 5.0, which may interfere with the compilation system.

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/283)


jarro2783:
Thanks, looks good.

PR #285, opened 2021-04-23
Fully compatible with GCC4.8 compilation system.

RonxBulld:
G++ 4.8's regular expression library is very bad, it can't parse and process regular expressions in code.After research, I decided to write a lexical analyzer by hand.

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/285)


jarro2783:
Do you have a use case for this? I never really intended this for GCC 4.8, and it's _very_ old now.

It looks like some of those commits duplicate what you've already done in the previous PR from you. Do you need to remove some of them?

jarro2783 (commented):
I think I would like to see this extra code encapsulated into something so it can be tested. Then the call site either wraps a single function call in `#ifdef `or even uses a template specialisation with the defined constant to call the right code.

RonxBulld:
> 
> 
> Do you have a use case for this? I never really intended this for GCC 4.8, and it's _very_ old now.
> 
> It looks like some of those commits duplicate what you've already done in the previous PR from you. Do you need to remove some of them?

For example: CentOS, some third-party forked systems, traditional industrial software, and non-mainstream chip architectures usually require an earlier version of the compiler. gcc-4.8 is the version number they often mention.

include/cxxopts.hpp
```diff
@@ -594,6 +607,34 @@ namespace cxxopts
     void
     integer_parser(const std::string& text, T& value)
     {
+#ifdef NO_REGEX
+      if (text.empty()) {
```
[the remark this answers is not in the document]
RonxBulld: > 
> 
> Can you separate this functionality out into some functions that can be unit tested?

I agree with your suggestion, and I will do it in the next work.

jarro2783:
are you able to add a gcc 4.8 build to the travis file so that we can make sure everything still works?

RonxBulld:
> 
> 
> are you able to add a gcc 4.8 build to the travis file so that we can make sure everything still works?

Done.

proposals-demo/CMakeLists.txt
```diff
@@ -0,0 +1 @@
+add_subdirectory(objectification)
```
jarro2783, line 1: what's all this for?

RonxBulld: I wanted to improve some ease of use, but I accidentally pushed it to github, and then it was automatically synchronized to the PR stream. How can I remove the commit from 491cad9 in the RP stream?

jarro2783:
You can do an interactive rebase to drop the commit, then just force push.

jarro2783 (commented):
Looks good, just that one symbol to change.

PR #286, opened 2021-05-05
Support options like `-j5`.

RonxBulld:
I suggest to support parameter formats like `-j5` to make the input look more GNU style.

I made some minor changes: if a short option is recognized and it requests a parameter, any characters after it are treated as the parameter of the option.

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/286)


jarro2783:
Can you write some tests?

RonxBulld:
Done.

PR #298, opened 2021-07-21
Added CPack packaging, pkg-config and CMake version files.

KOLANICH:
Made the path to stuff architecture-independent, since the library is header-only.

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/298)


jarro2783:
Looks good. I haven't actually used cpack. How do I run it just so that I can see what this does?

KOLANICH:
I am assumming that you are on Kubuntu/Debian and use `ninja` for building. For other targets the procedure is the similar.

1. `cmake-gui`
2. Set the variable `CPACK_GENERATOR` to `DEB`.
3. Set the prefix to `/usr`
4. configure and generate
5. `ninja package` would generate the deb package
6. Examine the package contents with the archiver. I use `Ark`, though it is an extremily bad archiver.
7. `sudo dpkg -i ./*.deb` would install it into the system
8. Then `find_package` and `pkg_check_modules` should be working in other CMake projects.

Issue #223, opened 2020-03-02
Add a print function for the option object


cyang-kth:
I think that a print() function can be useful to print out the configurations parsed by the user. The description can be reused during the printing.

```
cxxopts::Options options;
auto result = options.parse(argc, argv);
options.print(result);
```

jarro2783:
What sort of information would you like printed that would be different from what `help` prints?

cyang-kth:
@jarro2783 

The current help print like this
```
      --log_level arg         Log level (default: 3)
      --network arg           Network file name
      --id arg                Network id name (default: id)
```

When the user types 
```
--log_level 5 --network foo.txt --id gid
```

I would like to have
```
      --log_level arg         5
      --network arg           foo.txt
      --id arg                gid
```
I frequently encounter this need when I want to confirm the input argument is correct and record it in the log file.

cwreynolds:
For what it is worth, in the last project where I used `cxxopts`, I wrote code nearly identical to this. Before beginning a long period of computation, I had it list all the parameters it was using.

jarro2783:
Ok that makes. So you actually want it to print out the `ParseResult` object so you know what it has actually parsed.

Issue #126, opened 2018-08-29
Getting full arguments value set as string including default & implicit


m4saka:
Hi.

My application will have a lot of parameters, so I'd like to get the full list of arguments from cxxopts and save the set as a text file.

Is it possible to iterate all arguments with specified values including default and implicit ones with cxxopts?

`ParseResult::arguments()` enables to get an argument list as `std::vector<KeyValue>`, but it doesn't contain default values.
And `ParseResult` has unordered_map-like overloads but it cannot be used for iteration.
```cpp
auto result = options.parse(argc, argv);

// You can't do this
for (auto && argument : result)
{
    std::cout << argument.first << "=" << argument.second.as<std::string>() << std::endl;
}
```

jarro2783:
I can work on adding something like that to the interface.

PR #313, opened 2021-10-25
Add iterator (#126) and easily print-out (#223) to ParseResult

ldeng-ustc:
## Enhancements

Add `m_defaults` to record options used the default value.

Add `ParseResult::Iterator` to iterate all `KeyValue`s (including default values). (#126 )

Add `ParseResult::arguments_string` to easily print out all parsed options. (#223 )

Add `-l, --list` and `--range` options to `example.cpp` .

### Easily parsed values printing

Print all parsed options names and values by `arguments_string` :

```cpp
options..add_options()
// define some options here
;
auto result = options.parse(argc, argv);
std::cout << result.arguments_string() << std::endl;
```

You can run `example` with `-l` options to see the output.

```bash
$ ./src/example -l  -- input output p1,p2,p3
list = true
input = input
output = output
positional = p1,p2,p3
range = false (default)
compile = false (default)
tab-expansion = false (default)
help = false (default)
long-description = false (default)
apple = false (default)
option_that_is_too_long_for_the_help = false (default)
true = true (default)
bob = false (default)
output = a.out (default)
```

### Iterator and range-for

Use range-for to print all parsed values:

```cpp
for(const auto &kv: result)
{
  std::cout << kv.key() << " = " << kv.value() << std::endl;
}
```

Run `example` with options `-l` and `--range` to see the output:

```bash
$ ./src/example -l --range  -- input output p1,p2,p3
list = true
range = true
input = input
output = output
positional = p1,p2,p3
compile = false
tab-expansion = false
help = false
long-description = false
apple = false
option_that_is_too_long_for_the_help = false
true = true
bob = false
output = a.out
```

## Problems

* I have no idea about how to process the Unicode options.
* The output is not aligned.

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/313)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -1480,7 +1480,6 @@ namespace cxxopts
         ++m_iter;
         if(m_iter == m_pr->m_sequential.end())
         {
-          std::cout << "into defaults" << std::endl;
           m_iter = m_pr->m_defaults.begin();
           return *this;
         }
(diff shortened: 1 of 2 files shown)


include/cxxopts.hpp
```diff
@@ -1480,7 +1480,6 @@ namespace cxxopts
         ++m_iter;
         if(m_iter == m_pr->m_sequential.end())
         {
-          std::cout << "into defaults" << std::endl;
```
jarro2783: I'm not sure why this line was here, thanks.

Issue #317, opened 2021-12-03
arguments() contains only long_name of options


ldeng-ustc:
When use a short-only options name, `ParseResult::arguments.key()` will be empty string, because only long_name will be insert into m_sequential in `parse_option`:

```cpp
// line 2164~2178
inline
void
OptionParser::parse_option
(
  const std::shared_ptr<OptionDetails>& value,
  const std::string& /*name*/,
  const std::string& arg
)
{
  auto hash = value->hash();
  auto& result = m_parsed[hash];
  result.parse(value, arg);

  m_sequential.emplace_back(value->long_name(), arg); // empty key could be insert into m_sequential when option is short-only name.
}
```

Also, `arguments_string()` will output parse results with no name like (where I set short-only option `n=16` and `m=8`):
```bash
 = 16
 = 8
help = false (default)
long = false (default)
``` 

Possible solution:
Add `essential_name()` to `OptionsDetails`, which return long name, or short name when long name is empty. And use these as key in `m_sequential` and `m_defaults`

Maybe I will create a PR to solve these.

PR #318, opened 2021-12-03
Solve arguments() showing empty keys when exsiting short-only option

ldeng-ustc:
Solve #317

Before this PR:
```sh
$ ./example --list -x 1        
list = true
 = 1            # expect "x = 1" here
range = false (default)
compile = false (default)
...
```

After:
```sh
$ ./example --list -x 1        
list = true
x = 1          # OK
range = false (default)
compile = false (default)
```

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/318)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -1287,6 +1287,13 @@ namespace cxxopts
       return m_long;
     }
 
+    CXXOPTS_NODISCARD
+    const std::string&
+    essential_name() const
+    {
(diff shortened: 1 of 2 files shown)


jarro2783:
Thanks, can you write a unit test?

ldeng-ustc:
I add check on `arguments()` key and value in short option unit test, and write a new unit test for iterator.

jarro2783:
Thanks for your PR.

Issue #327, opened 2022-04-05
Derive `option_has_no_value_exception` from `OptionParseException`


MarcoKoch:
Currently, `option_has_no_value_exception` is derived directly from `OptionException`, whereas other exceptions that may occur when examining a `ParseResult` are all derived from `OptionParseException`. This requires an extra `catch` block for handling bad arguments.

MarcoKoch:
As discussed in the PR, this suggestion was based on a wrong assumption on my part.

PR #328, opened 2022-04-05
Derive `option_has_no_value_exception` from `OptionParseException`

MarcoKoch:
Implements #327.

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/328)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -491,11 +491,11 @@ namespace cxxopts
     }
   };
 
-  class option_has_no_value_exception : public OptionException
+  class option_has_no_value_exception : public OptionParseException
   {
     public:
     explicit option_has_no_value_exception(const std::string& option)
-    : OptionException(
+    : OptionParseException(
         !option.empty() ?
         ("Option " + LQUOTE + option + RQUOTE + " has no value") :
         "Option has no value")


jarro2783:
This isn't a parse error, it is a usage error. So there is a reason that the two are separated.

MarcoKoch:
Seems like I had a misunderstanding about what that exception exactly means.

Just to clarify that I understand your reasoning correctly:
I came across this exception when trying to read the value of an optional positional argument, which wasn't supplied by the user. Am I right that I'm supposed to use `ParseResult::count()` to check whether a value is present before trying to read it with `OptionValue::as()`? So there are actually three categories of exceptions:
1. Exceptions from invalid option specifications (`OptionSpecException`)
2. Exceptions from invalid arguments supplied by the user (`OptionParseException`)
3. Exceptions from trying to access the results of a parse in an invalid manner (which `option_has_no_value_exception` falls into)

jarro2783:
Yes that's correct. I think all of those should derive from `OptionException`, so if you want to catch everything in one go you can just catch that instead.

MarcoKoch:
Adding an extra `catch` is fine now I understand the difference. I probably was just a bit confused because the README only mentions two categories of exceptions.
Maybe it would also make sense to add a third base class to reflect the conceptual difference in code, even though `option_has_no_value_exception` would be the only one that derives from it at the moment.
If you are okay with that, I will update my pull request accordingly.

Issue #334, opened 2022-06-30
Clarify docs on positional arguments


matthew-limbinar:
More than one person using cxxopts has been confused about how to use positional arguments, and the docs don't help much. Please clarify.

PR #335, opened 2022-06-30
Issue 334: Clarify docs on positional arguments

matthew-limbinar:
This would resolve https://github.com/jarro2783/cxxopts/issues/334.

This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/335)


jarro2783:
That is much clearer, thanks for putting the time into to improve the documentation.

PR #342, opened 2022-07-07
Fixes #338: Split the library into multiple header files

### LINKED ISSUE #338: Split cxxopts.hpp into multiple files
One can observe a rather straightforward decomposition of `cxxopts.hpp` into several parts:

*    non-option-specific string-related code
*    Exception-related code
*    the values namespace and the Value class
*    Options and classes with Option in their name
*    The ParseResult class and some related code.
*    Help-related definitions and code.

And the above is also a topological order of dependencies (except for some implementations which cxxopts.hpp has at the bottom).

So, how about we split the huge single file into a file for each of these? With `cxxopts.hpp` remaining as an includer of all the others?

This will:

* Improve source code readability
* Make it easier to merge changes
* Make the revision history easier to understand (when looking at the affected files)

What do you think, @jarro2783 ?

eyalroz:
Fixes #338: Split the library into multiple header files, mostly one per class.

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/342)


eyalroz:
Still thinking about separating the help-related classes, or perhaps the OptionAdder. Although - OptionAdder could simply be made an inner class of Options.

PR #347, opened 2022-07-08
Fixes #345, fixes #346: Exception code tweaks

### LINKED ISSUE #345: Put exceptions in a sub-namespace rather than suffixing them with "_exception"
Why use `option_has_no_value_exception` rather than `exception::option_has_no_value`?

### LINKED ISSUE #346: option_required_exception and option_not_present_exception sound like the same thing
Why is there an `option_required_exception` but also an `option_not_present_exception`? It sounds like they're about the same situation.

Looking at the code, `option_not_present` is when you're looking up an option in a ParseResult instance. Well, in that case, I'd rename: `required_option_missing_exception` and `requested_option_missing` / `requested_option_present`.

eyalroz:
This puts exceptions in their own sub-namespace and renames some of them accordingly.

Note that I think there might be an over-abundance of exception classes in the library, but so be it...


jarro2783:
This is good, but I think it needs rebasing after the namespace fixes.

eyalroz:
Yes, I'll take care of that soon.

eyalroz:
@jarro2783 : Done, and also had to make a few more changes regarding namespace indentation which slipped by me in PR #350.

eyalroz:
@jarro2783 : After this, let's go for #349 .

eyalroz:
So, somehow after the merge, the exceptions namespace itself is indented. Will submit another PR.

Issue #240, opened 2020-07-07
Support option name aliases / multiple long options


eyalroz:
(This is a feature I added to boost::program_options a couple of years back.)

Right now, if you name an option `"x,long-name-of-x,another-long-name-of-x"` - you get an error: 
```
terminate called after throwing an instance of 'cxxopts::invalid_option_format_error'
  what():  Invalid option format 'x,long-name-of-x,another-long-name-of-x'
```
instead, cxxopts should support having multiple aliases for the (first and primary) name of the option - all of which would be accepted on the command-line just as though the primary name were used.

jarro2783:
I can't think of anything, a short and a long name are basically already an alias for each other, so I think the syntax just needs to be extended. Go for it.

KappaDistributive:
Sorry for the slow reply -- live happened.

I've taken a closer look at the code now and thought about how to implement it. AFAIK, there are a few viable options, ranked in increasing order of intrusiveness:

- We could overload `Options::add_option` once more and allow it to receive an array of long names and then adapt `OptionAdder::operator()` and the relevant RegEx accordingly. (The only option that is 100% backward compatible)
- We could do away with the assumption that long names consist of only one alias and replace the current implementation of `Options::add_option` and `OptionAdder::operator()`. (`OptionAdder::operator()` would remain backward compatible but `Options::add_option` would not.)
- We could add an additional layer of abstraction and create a bunch of `Command` (name tbd) entities that we split the initial string into and then handle both short and long names uniformly. (`OptionAdder::operator()` would remain backward compatible but `Options::add_option` would not.)

Since I don't know this package and its users well enough to make this call, I require your expertise to make this decision. Once we've narrowed it down, I'll create an initial implementation that we can discuss in detail.

eyalroz:
I think that the minimal change in terms of user-facing functions is simply being willing to parse "x,foo,bar" and not just "x,foo":
```
    options.add_options()
        ("x,foo,bar", "Frobincate the bar", cxxopts::value<bool>()->default_value("true"));
```
That's your second option IIUC.

But it doesn't have to be just one way of specifying multiple aliases. And I like the first option as well, i.e. I'd also like it if I could write:
```
    options.add_options()
        ({'x', "foo", "bar" }, "Frobincate the bar", cxxopts::value<bool>()->default_value("true"));
```
and for that to work.

KappaDistributive:
[...]

PR #349, opened 2022-07-09
Fix issue #240: Multiple long option names / aliases

eyalroz:
**Edit 2:** This PR is finalized, now with some testing

* We now use a vector of long option names instead of a single name
* When specifying an option, you can provide multiple names separated by commas, at most one of which may have a length of 1 (not necessarily the first specified name). The length-1 name is the single-hyphen switch (the "short name").
* Hashing uses the first long name
* Option help currently only uses the first long name. Not sure if that should change.
* Utilizing this feature in unit tests in `test/options.cpp` and also in `src/example.cpp`.

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/349)


jarro2783:
This looks like a good approach. Given the way short and long names are basically an alias already, I think this can mostly be reused as you have done to just make more aliases.

include/cxxopts.hpp
```diff
@@ -761,19 +764,40 @@ namespace cxxopts
...
         }
 
-        const std::string& short_sw = result[2];
-        const std::string& long_sw = result[3];
+        OptionNames split_names;
+        std::string::size_type token_start_pos = 0;
+        auto length = text.length();
 
-        return std::pair<std::string, std::string>(short_sw, long_sw);
+        // TODO: Try avoiding having to tokenize ourselves
```
jarro2783, line 781: `std::regex_iterator`?

eyalroz: std::token_regex_iterator will do, apparently. Thanks for the idea.

include/cxxopts.hpp
```diff
@@ -755,19 +778,23 @@ inline bool IsFalseText(const std::string &text)
...
-inline std::pair<std::string, std::string> SplitSwitchDef(const std::string &text)
+// Gets the option names specified via a single, comma-separated string,
+// and returns the separate, space-discarded, non-empty names
+// (without considering which or how many are single-character)
+inline OptionNames split_option_names(const std::string &text)
 {
-  std::match_results<const char*> result;
-  std::regex_match(text.c_str(), result, option_specifier);
-  if (result.empty())
+  if (not std::regex_match(text.c_str(), option_specifier))
```
jarro2783, line 786: can you use `!` instead of `not`

include/cxxopts.hpp
```diff
@@ -755,19 +778,23 @@ inline bool IsFalseText(const std::string &text)
...
   {
     throw_or_mimic<exceptions::invalid_option_format>(text);
   }
 
-  const std::string& short_sw = result[2];
-  const std::string& long_sw = result[3];
+  OptionNames split_names;
 
-  return std::pair<std::string, std::string>(short_sw, long_sw);
+  constexpr const int use_non_matches { -1 };
```
jarro2783, line 793: I think the extra `const` is redundant.

eyalroz: @jarro2783 : Only in C++11. If you compile with C++14 or later, `const` isn't implied. I don't mind very much, the only reason I have this is to avoid using `-1` as a "magic number".

include/cxxopts.hpp
```diff
@@ -2111,36 +2169,34 @@ OptionAdder::operator()
...
-  std::string short_sw, long_sw;
-  std::tie(short_sw, long_sw) = values::parser_tool::SplitSwitchDef(opts);
-
-  if (!short_sw.length() && !long_sw.length())
-  {
-    throw_or_mimic<exceptions::invalid_option_format>(opts);
+  OptionNames all_option_names = values::parser_tool::split_option_names(opts);
+    // Note: All names will be non-empty!
+  std::string short_sw {""};
+  OptionNames long_option_names = all_option_names;
```
jarro2783, line 2175: do you need this extra copy? It looks like you don't use `all_option_names` again.

include/cxxopts.hpp
eyalroz: I could... although personally I've long switched to only using `and`, `or` and `not` to make my code more natural-language like.

include/cxxopts.hpp
eyalroz: I don't. I just wanted to clarify this is no longer `all_option_names`.  Ideally, a compiler could optimize this away; but I suppose I could use `std::move`. On the other hand, there are soooo many string copies in cxxopts - it could really stand to use `string_view` everywhere.

include/cxxopts.hpp
```diff
@@ -2111,36 +2169,34 @@ OptionAdder::operator()
...
+  );
+  if (first_length_1_name_it != long_option_names.cend()) {
+    auto have_second_length_1_name = std::any_of(first_length_1_name_it + 1, long_option_names.cend(),
+      [&](const std::string& name) { return name.length() == 1; }
+    );
+    if (have_second_length_1_name) {
+      throw_or_mimic<exceptions::invalid_option_format>(opts);
+    }
+    short_sw = *first_length_1_name_it;
+    long_option_names.erase(first_length_1_name_it);
```
jarro2783, line 2187: This seems a bit complicated. Are you basically trying to find one short name and then leave multiple long names?
I wonder if a simple loop would just be clearer. Or maybe you could use `remove_if` and then check if the remaining list is greater than 1.

jarro2783 (commented):
Overall looking pretty good, just a few nitpick comments.

include/cxxopts.hpp
jarro2783: Ok I didn't realise that, this is fine then.

include/cxxopts.hpp
jarro2783: Well remove_if doesn't actually remove it, it only moves everything to the end, so I think this is what you want.

jarro2783: Oh I just read that the values at the end have unspecified value, so they might get destroyed. In that case partition is probably the way to go.

eyalroz: Wait, actually, no, that's not quite what happens. If I `std::remove_if()` on a vector, the moved strings at the end don't keep their original values. So, that won't work.

include/cxxopts.hpp
jarro2783: Actually I think you've misread the statement about const.
> A constexpr specifier used in an object declaration [or non-static member function (until C++14)] implies const

It is only non-static member functions that were dropped from implying const. So this extra `const` is redundant.

jarro2783:
This is good, thanks for the effort. Can you just update that `const` line, and can you also add an entry into the changelog?

include/cxxopts.hpp
eyalroz: Well, that's counter-intuitive, but ok. Dropped the const.

eyalroz:
@jarro2783 : I already have updated the const line...

PR #355, opened 2022-07-25
Add option to display usage

yassirnajmaoui:
Make the usage displaying optional.

This was necessary for one of our projects since we had a custom usage for the program; we expected input from the standard output.

---
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/355)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -2725,10 +2725,13 @@ Options::generate_all_groups_help(String& result) const
 
 inline
 std::string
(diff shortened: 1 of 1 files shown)


jarro2783:
Looks like there are build failures.

yassirnajmaoui:
> Looks like there are build failures.

You're right. I forgot some definitions.
I fixed it now, please tell me if something else breaks.

Issue #364, opened 2022-09-01
When using Chinese Debug mode in Visual Studio, the assertion fails in std::isblank


66maer:
![image](https://user-images.githubusercontent.com/10154859/187870478-92f84c5e-2f4d-47cb-90c9-d11bf57df638.png)

![image](https://user-images.githubusercontent.com/10154859/187870760-a2e6b0d3-44ae-4e43-8fba-d912c9eba9c7.png)

```c++
extern "C" int __cdecl _chvalidator(int const c, int const mask)
{
    _ASSERTE(c >= -1 && c <= 255);
    return _chvalidator_l(nullptr, c, mask);
}
```

jarro2783:
Fixed by #365

hillmich:
This PR breaks the build on our CI on MacOS (see [corresponding run](https://github.com/cda-tum/ddsim/runs/8264664336?check_suite_focus=true) on cda-tum/ddsim#123, error also stated below). It seems to me this is not caused by our code. Would you be wiling to have a look?

```
...
In file included from /Users/runner/work/ddsim/ddsim/apps/simple.cpp:10:
/Users/runner/work/ddsim/ddsim/extern/cxxopts/include/cxxopts.hpp:2093:33: error: too many arguments to function call, expected single argument '_c', have 2 arguments
    if (std::isblank(*previous, std::locale::classic()))
        ~~~~~~~~~~~~            ^~~~~~~~~~~~~~~~~~~~~~
/Applications/Xcode_13.2.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.1.sdk/usr/include/_ctype.h:224:1: note: 'isblank' declared here
isblank(int _c)
^
In file included from /Users/runner/work/ddsim/ddsim/apps/simple.cpp:10:
/Users/runner/work/ddsim/ddsim/extern/cxxopts/include/cxxopts.hpp:2098:33: error: too many arguments to function call, expected single argument '_c', have 2 arguments
    if (!std::isblank(*current, std::locale::classic()))
         ~~~~~~~~~~~~           ^~~~~~~~~~~~~~~~~~~~~~
/Applications/Xcode_13.2.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.1.sdk/usr/include/_ctype.h:224:1: note: 'isblank' declared here
isblank(int _c)
^
2 errors generated.
```

jarro2783:
[...]

PR #368, opened 2022-09-10
Add missing `#include<locale>`

hillmich:
The recent change in #364 used `std::isblank(charT, locale)` from the header `locale` but did not include this header. Instead it included `cctype` which only provides `std::isblank(int)`. This caused a break for Xcode (see discussion in #364).

This tiny PR adds the required `#include<locale>`. (It might be worthwhile to switch all functions from [`cctype`](https://en.cppreference.com/w/cpp/header/cctype) to [`locale`](https://en.cppreference.com/w/cpp/header/locale).)

- - -
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/368)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -33,6 +33,7 @@ THE SOFTWARE.
 #include <exception>
 #include <limits>
 #include <list>
+#include <locale>
 #include <map>
 #include <memory>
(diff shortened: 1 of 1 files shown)


jarro2783:
Well spotted, thanks.

PR #370, opened 2022-09-13
Extend CI and small fixes

### LINKED ISSUE #369: Building on MacOS fails
As of now, MacOS still fails (PR #368 did not fix it, sorry I messed that up).

I forked the repo and made quite a lot of changes trying to make `std::isblank` work on [hillmich/cxxopts](https://github.com/hillmich/cxxopts). The main change is to replace `std::isblank` by explicitly checking for space and tab as the [standard](https://en.cppreference.com/w/cpp/string/byte/isblank#See_also) dictates for the C locale. This fork also includes

- CI on MacOS and Windows
- Assume header `regexp` is always there since cxxopts requires c++11 or higher (and eliminate a large chunk of code)
- Remove some unused headers
- Get around requiring `cstring`
- Add `std::` to instances of `size_t`

@jarro2783 are you interested in these changes and would you accept them as PR?

hillmich:
This PR includes quite a few changes, mostly the extended CI now also builds and tests for MacOS as well as Windows ~~and the `<regexp>` header is now always assumed to be available since cxxopts requires c++11~~.

Smaller changes in this PR are:

- Remove reliance on `std::isblank(charT, locale)` so builds on MacOS work again. (fixes #369 )
- CI runs on `master` as well as `main` in case the default branch is changed in the future
- Remove some unused headers
- Add `std::` to instances of `size_t`

- - -
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/370)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -27,24 +27,33 @@ THE SOFTWARE.
 #ifndef CXXOPTS_HPP_INCLUDED
 #define CXXOPTS_HPP_INCLUDED
 
-#include <algorithm>
+#include <cstring>
 #include <exception>
 #include <limits>
 #include <initializer_list>
 #include <map>
 #include <memory>
-#include <regex>
 #include <sstream>
 #include <string>
 #include <unordered_map>
 #include <unordered_set>
 #include <utility>
 #include <vector>
+#include <algorithm>
 
 #ifdef CXXOPTS_NO_EXCEPTIONS
 #include <iostream>
 #endif
 
+#if defined(__GNUC__) && !defined(__clang__)
+#  if (__GNUC__ * 10 + __GNUC_MINOR__) < 49
+#    define CXXOPTS_NO_REGEX true
+#  endif
+#endif
+
+#ifndef CXXOPTS_NO_REGEX
+#  include <regex>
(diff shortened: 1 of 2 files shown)


include/cxxopts.hpp
```diff
@@ -582,168 +570,6 @@ struct ArguDesc {
   std::string value     = "";
 };
 
-#ifdef CXXOPTS_NO_REGEX
```
jarro2783, line 585: this code was added not that long ago, I don't think we should remove it yet since it is providing support for fairly old compilers.
What was broken for you that would make you want to get rid of it?

hillmich:
It's not broken in that sense, I just figured since cxxopts is using C++11 features anyway (e.g., `std::shared_ptr`, type alias with `using`, `constexpr` specifier, and `std::forward`) we might as well assume full compliance with the standard and require `<regexp>` to be available.
What do you think?

jarro2783:
I'd prefer you to keep that code in for now given that it was supporting an old compiler that was actually being used still.

hillmich:
Okay, I reverted the corresponding commit

jarro2783:
Looks good, thanks.

PR #377, opened 2022-11-04
Fix an additional space in help generated

soumeh01:
### Problem: 
In the scenario when command line args don't have any options and have only positional args. The help generated shows an extra space when we try to set custom help to `empty string` in order to overwrite the default `[OPTION...]` string in the help.

```
cxxopts::Options options{"test", ""};
std::vector<std::string> positional;
options
  .positional_help("<posArg1>...<posArgN>")
  .custom_help("")
  .add_options()
    ("positional", "", cxxopts::value<std::vector<std::string>>(positional));

options.parse_positional({"positional"});
std::vector<const char*> arg = {"pos1", "pos2", "pos3"};
auto r = o.parse(arg.size(), arg.data());
cout<<options.help();
```

### Current generated help
```
Usage:
  test  <posArg1>...<posArgN>
```
Note: there are 2 spaces between `test` and `<posArg1>`
### Expected help
```
Usage:
  test <posArg1>...<posArgN>
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/377)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -2742,7 +2742,12 @@ Options::help(const std::vector<std::string>& help_groups, bool print_usage) con
   String result = m_help_string;
   if(print_usage)
   {
-    result+= "\nUsage:\n  " + toLocalString(m_program) + " " + toLocalString(m_custom_help);
+    result+= "\nUsage:\n  " + toLocalString(m_program);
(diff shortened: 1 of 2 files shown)


jarro2783:
Thanks.

PR #386, opened 2023-01-28
fuzz: Adds fuzz test

silvergasp:
Adds a fuzz test harness. I run the fuzzer for about 12hrs on 8 cores and didn't find any bugs, which is quite impressive :) So you've done well with your library. I thought I'd open a pull request with the hopes that the fuzzer can be integrated into the CI.

- - -
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/386)

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -20,11 +20,9 @@ jobs:
   build-ubuntu:
     strategy:
       matrix:
-        os: [ ubuntu-18.04, ubuntu-20.04, ubuntu-22.04 ]
+        os: [ ubuntu-20.04, ubuntu-22.04 ]
         compiler: [ g++-9, g++-10, clang++ ]
-        include:
-          - os: ubuntu-18.04
-            compiler: g++-7
+
     name: Build and Test on Ubuntu
     runs-on: ${{matrix.os}}
     steps:
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -6,3 +6,11 @@ cc_library(
     strip_include_prefix = "include",
     visibility = ["//visibility:public"],
 )
+
+load("@rules_fuzzing//fuzzing:cc_defs.bzl", "cc_fuzz_test")
+
+cc_fuzz_test(
+    name = "cxxopts_fuzz_test",
+    srcs = ["test/fuzz.cpp"],
+    deps = [":cxxopts"],
+)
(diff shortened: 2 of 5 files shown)


jarro2783:
Is this generating random input to the options parser and then checking that it doesn't segfault? Would you also use fuzzing to somehow check that a range of input is processed correctly (like integers being parsed correctly) or is it more often just to check that the program doesn't crash on weird input?

Can you integrate this with the CMake build as well?

silvergasp:
> Is this generating random input to the options parser and then checking that it doesn't segfault? Would you also use fuzzing to somehow check that a range of input is processed correctly (like integers being parsed correctly) or is it more often just to check that the program doesn't crash on weird input?

Yeah, that's pretty much it, although there are some heuristics to guide the process. For example, if a random input achieves good code coverage, instead of trying again with new random data, it'll randomly mutate the original data. There is also the option to combine it with various sanitizers. If you aren't familiar with sanitizers you can think of these as runtime "bug detectors". Here's a non-exhaustive list of sanitizers that are supported; 
- Address sanitizer: Detects buffer overflows, amongst other bugs
- Undefined behaviour sanitiser: Among other things detects integer overflows/divide by zero's.
- etc. (Search for sanitizer [here](https://clang.llvm.org/docs/index.html) for more details).

> Can you integrate this with the CMake build as well?

Yeah sure, I rarely use Cmake so that might take me a bit to work out what I need to do :)

silvergasp:
RE:
> range of input is processed correctly (like integers being parsed correctly)

This is very much supported. Though the reason I didn't include it in this fuzzing harness is that most of that seems to happen using the `>>` operator which is either provided by the standard library or user provided. So I may be wrong here, but the assumption that I made was that the parsing of string->variable, wasn't really covered by this library. More so you provide a convenient interface to implement this on your own.

silvergasp:
Some notes for recent changes:
- I removed ubuntu 18.04 from the CI as it was causing problems [and is depracated](https://github.blog/changelog/2022-08-09-github-actions-the-ubuntu-18-04-actions-runner-image-is-being-deprecated-and-will-be-removed-by-12-1-22/)
- I enabled the fuzz tests to run on linux+clang as it's not supported with gcc and I couldn't get it working out of the box with macos. Libfuzzer is not supported anywhere else.

Sorry for the CI spam :)

silvergasp:
As requested I also added the Cmake support

jarro2783:
This looks good, thanks.

PR #392, opened 2023-02-18
feat(fuzz): Adds trigger for fuzzing in the CI

silvergasp:
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/392)

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,24 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+ Fuzzing:
+   runs-on: ubuntu-latest
+   steps:
+   - name: Build Fuzzers
+     id: build
+     uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+     with:
+       oss-fuzz-project-name: 'cxxopts'
+       language: c++
+   - name: Run Fuzzers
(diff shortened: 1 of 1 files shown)


.github/workflows/cifuzz.yml
```diff
@@ -0,0 +1,24 @@
...
+     uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+     with:
+       oss-fuzz-project-name: 'cxxopts'
+       language: c++
+   - name: Run Fuzzers
+     uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+     with:
+       oss-fuzz-project-name: 'cxxopts'
+       language: c++
+       fuzz-seconds: 600
```
jarro2783, line 18: I think for a PR we could run it for less time since it will mean waiting for a build for 10 minutes. Maybe you could do a branch build that is longer too.

silvergasp: No worries, how does 5min sound? From my experience, anything less than that will add very little value. Could you expand on what you mean by branch build?

jarro2783: 5 minutes sounds good. I meant building on a push to master, but I just realised that it runs daily with the integration you set up, so I don't think there is any need for that.

silvergasp: Done

PR #395, opened 2023-03-08
Output same exception messages for both Linux and Windows

nigels-com:
While it is indeed pleasing to see UTF-8 output on Linux, we strongly prefer cross-platform consistency to help with testing.

If this approach  is too heavy-handed, I could refactor with some sort of opt-out for this.

- - -
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/395)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -337,13 +337,8 @@ empty(const std::string& s)
 namespace cxxopts {
 
 namespace {
-#ifdef _WIN32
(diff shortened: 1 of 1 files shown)


jarro2783:
This was done originally because Windows couldn't display these characters correctly. Can you check that the output looks sensible on Windows?

nigels-com:
Yes, so the change to go for lowest common denominator - ASCII.

jarro2783:
I guess this is fine. Maybe I'll build in an opt out later.

Issue #405, opened 2023-05-25
std::bad_alloc during dlopen on integer_pattern initialization


mxmlnkn:
Hello,

I'm using cxxopts v2.2.1 for argument parsing in [indexed_bzip2](https://github.com/mxmlnkn/indexed_bzip2) and got a very weird [user-reported issue](https://github.com/mxmlnkn/indexed_bzip2/issues/14#issuecomment-1562426673).

```pascal
#2  0x00007fcac0dc77ec in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6                                                                                                   [150/22615]
#3  0x00007fcac0dd2966 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x00007fcac0dd29d1 in std::terminate() () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#5  0x00007fcac0dd2c65 in __cxa_throw () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x00007fcac0dc742a in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#7  0x00007fcac0e607b2 in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_M_assign(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocato
r<char> > const&) () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#8  0x00007fc9d5eb2c65 in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::assign (__str="", this=0x7ffce48d9540)
    at /opt/rh/gcc-toolset-11/root/usr/include/c++/11/bits/basic_string.h:1387
#9  std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::operator= (__str="", this=0x7ffce48d9540)
    at /opt/rh/gcc-toolset-11/root/usr/include/c++/11/bits/basic_string.h:681
#10 std::__detail::_Compiler<std::__cxx11::regex_traits<char> >::_M_match_token (token=std::__detail::_ScannerBase::_S_token_subexpr_begin, this=0x7ffce48d9430)
    at /opt/rh/gcc-toolset-11/root/usr/include/c++/11/bits/regex_compiler.tcc:593
[...]

PR #406, opened 2023-05-29
Avoid std::regex initialization during dlopen

mxmlnkn:
Suggested workaround for #405. Unfortunately, I have no minimal reproducer to test it against.

- - -
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/406)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -55,8 +55,8 @@ THE SOFTWARE.
 #define CXXOPTS_LINKONCE_CONST	__declspec(selectany) extern
 #define CXXOPTS_LINKONCE		__declspec(selectany) extern
 #else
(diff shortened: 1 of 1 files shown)


mxmlnkn:
I have already built a new release (indexed_bzip2 1.5.0) using this forked version of cxxopt and asked @ozancaglayan to try it out. I am yet to hear back.

jarro2783:
This seems like a good change anyway, but I would like to hear back if it solves the original problem.

mxmlnkn:
> Are you able to get your user to try it with this fix to at least see if it fixes their actual problem.

It (or some other change between indexed_bzip2 1.4.0 and 1.5.0) did fix the problem.

PR #411, opened 2023-11-23
Minor typo in README.md

Blake-Madden:
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/411)

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,8 +193,8 @@ therefore, `-o false` does not work.
 
 ## `std::vector<T>` values
 
-Parsing of list of values in form of a `std::vector<T>` is also supported, as long as `T`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -193,7 +193,7 @@ therefore, `-o false` does not work.
 
 ## `std::vector<T>` values
 
-Parsing of list of values in form of an `std::vector<T>` is also supported, as long as `T`
+Parsing of list of values in form of a `std::vector<T>` is also supported, as long as `T`
```
jarro2783, line 196: Thanks for the fix, but I'm thinking that this is a terrible sentence and should probably be rewritten more to something like
> Parsing a list of values into a `std::vector<T>` is ...

Blake-Madden: OK, I reworded that sentence some more. I also say "define" for a define symbol instead of "definition."

Issue #363, opened 2022-08-30
Use first, not last, long name in usage message


eyalroz:
After the fix for #240 is in, the library must choose a long option for the program usage information.

At the moment, it seems to be choosing the last one specified. Instead, it should choose the first.

jarro2783:
I'll take a look at that, I hope it's not putting it into an unordered map or something.

lazysegtree:
With this https://github.com/jarro2783/cxxopts/pull/412, we will rightfully print all long names. So the ambiguity of which name is picked won't be there.

eyalroz:
_All_ names? I'm not sure that's such a good idea; will comment there.

But regardless - if you do print all names, this bug becomes "make sure the first name printed is the first long name on the list."

lazysegtree:
@eyalroz #481 should fix the behavior for now. Before we reach any conclusion on #412

PR #412, opened 2023-11-30
Print long names in help option

amirsojoodi:
Related to #240, #363, and #349

Tested w/wo small names

- - -
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/412)

### DIFF
--- a/test/options.cpp
+++ b/test/options.cpp
@@ -1191,3 +1191,15 @@ TEST_CASE("No Options help", "[options]")
   CHECK_NOTHROW(options.parse(argc, argv));
   CHECK(options.help().find("test <posArg1>...<posArgN>") != std::string::npos);
 }
+
+TEST_CASE("Help shows all aliases", "[help]")
+{
+  cxxopts::Options options("Print all aliases", " - ensure help lists every long name");
+
+  options.add_options()
+    ("v,verify,ver", "Verification option", cxxopts::value<std::string>());
+
+  const auto help = options.help();
+
+  CHECK(help.find("-v, --ver, --verify arg") != std::string::npos);
+}


lazysegtree:
@amirsojoodi Are you still working on this?

amirsojoodi:
@lazysegtree I've forgotten about this, too. I'll take a look

amirsojoodi:
I have an option added like this:

```cpp
    options.add_options()("v,verify,ver", help::verify, cxxopts::value<string>());
```

`options.help()`, before the change:

```txt
  -v, --verify arg    Verification option can be used to specify one type of verification 
                      that is performed after each transfer. The options are: none, 
                      eager, lazy. The default option is none.
```

and after the change:

```txt
  -v, --ver, --verify arg       Verification option can be used to specify one type of 
                                verification that is performed after each transfer. The 
                                options are: none, eager, lazy. The default option is 
                                none.
```

This is just an example. I have many other options like:

```cpp
    options.add_options()("S,src_device,src", "Source Device ID", cxxopts::value<int>());
    options.add_options()("D,dst_device,dst", "Destination Device ID", cxxopts::value<int>());
```

lazysegtree:
@jarro2783 When you are available, can you take a look at the changes? Your comments are addressed.

lazysegtree:
@amirsojoodi the CI was fixed and now passes, can you rebase with latest master ?

Edit - In case there is no response within 2 weeks, I think we will take these changes via a different PR.

eyalroz:
This is a bad idea in the general case. Why? Because there may very well be many long names. I have apps with 6 or 8 different long names for the same options - and they're not just 3 or 6 or 8 characters long, but 10 or 20 characters. I really don't want to spew out all of the slight variations on phrasing, onto the terminal with the usage information, just because I've decided to be flexible.

Either there is some control on which long names get printed, or only the first of them should get printed.

lazysegtree:
I don't have a strong opinion on this, but I agree with @eyalroz 
Only if there is a valid use case that needs all options printed, then we can have a configuration option for that. 

But maybe we shouldn't print them all by default.

@jarro2783 What do you think? (1) Should we print all long options by default? (2) Or print only the first? (3) Or allow a configurable option to print all?

eyalroz:
Some Possibilities for configurable control regarding long option printing:

* Method/function for setting the number of long options which get printed, after defining the options.
* An extra argument/potential argument for the number of long options which get printed, as part of the option definition.
* Separation of the long options into two sets: one which gets printed and one which doesn't. This can be done within the single string of option names, or by taking a container/initializer list of strings of the first kind and of the second kind.

The first is probably the easiest to implement and the least disruptive of existing workflows. But I don't have a strong preference.

jarro2783:
I agree this might not be good in some cases. Maybe if it was configurable it would be ok.

Another option is to have some heuristic about maximum length of options printed.

amirsojoodi:
Sorry for the delay in getting back to this.
Unfortunately, I cannot work on this PR anymore due to my new employment agreement, unless I have it approved by the compliance team. It will take a long time, so I have to delegate this to someone else. Sorry about that.

eyalroz:
@amirsojoodi : Oh, did you sign an agreement with one of those IP annexes which says "anything you work on while employed by us belongs to us, even if it isn't part of company business?" ... those are terrible. I've lost a potential job at least once because I insisted on amending that part.

... although, now that I think about it - if you like, perhaps one us (myself, @jarro2783 ) could get in touch with whoever is responsible, at your new employer, to petition them to approve this. Sometimes it helps if they don't perceive it as "our employer asking for a favor which we don't have to grant" but some kind of external party. Just a thought.

amirsojoodi:
@eyalroz Yep. Exactly. They have already rejected my request to work on a few other projects, after taking 3 months for "reviewing" my request. 

You made a thoughtful suggestion and I appreciate that. 

Tbh, my new job is important for me, and I think I don't want to bug them for now. But again, thank you.

jarro2783:
I'll see if I can do something quick to get this one over the line.

PR #414, opened 2023-12-20
Small spelling fixes

melroy89:
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/414)


jarro2783 (commented):
This code comes from an external library "catch" that I have just pasted into the repository.. I will update it and check if these have been fixed.

jarro2783:
Looks like these aren't in the new version.

PR #416, opened 2024-01-13
Ordered vector of group names to preserve order of groups

nigels-com:
Thanks for cxxopts, it's fits nicely with our needs.

I experimented with the grouping support today.  I noticed that the order of the groups isn't insertion order - it's based on the std::map comparator.  So I felt that we could add a std::vector of the group names here in the order of insertion.  Two benefits:

* The order of groups in the C++ code (insertion order) will match the `-h` output.
* The order of groups in the output will be deterministic and predictable, even across platforms and toolchains.

For consideration.

- - -
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/416)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -2033,6 +2033,7 @@ class Options
   std::unordered_set<std::string> m_positional_set{};
 
   //mapping from groups to help options
+  std::vector<std::string> m_group{};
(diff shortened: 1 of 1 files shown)


jarro2783:
Thanks, looks good.

PR #417, opened 2024-01-17
Fix overflow error in integer_parser (fixes #290)

### LINKED ISSUE #290: out of range argument not detected
An `unsigned long long` parameter throws a  `cxxopts::argument_incorrect_type` when called with `ULLONG_MAX + 1`  as expected
```
terminate called after throwing an instance of 'cxxopts::argument_incorrect_type'
  what():  Argument ‘18446744073709551616’ failed to parse
```
but cxxopts parses happily a much larger value `28446744073709551616` `(ULLONG_MAX + 1e19)` incorrectly and sets the value of the parameter to `1e19 - 1`.

I feel like both of those values should be refused. Am I doing something wrong or is this a bug related to the internal working of cxxopts?

ololuki:
Checking for overflow should be done before integer overflows.

There are two checks:
- (result > limit / base) is used for limits greater than rounded up to base, e.g. for 65535 it will activate for 65540 and higher.
- (result * base > limit - digit) is used for limit+1 to limit+n below next base rounded number, e.g. 65536 up to 65539.

fixes #290

- - -
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/417)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -27,6 +27,7 @@ THE SOFTWARE.
 #ifndef CXXOPTS_HPP_INCLUDED
 #define CXXOPTS_HPP_INCLUDED
 
+#include <cstdlib>
 #include <cstring>
 #include <exception>
 #include <limits>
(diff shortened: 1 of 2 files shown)


jarro2783:
Thanks for looking at this and fixing it for me.

PR #421, opened 2024-02-23
Extend cxxopts API to support std::optional non-throwing query of values

nigels-com:
We have some internal usage of `std::optional` that can integrate very tidily with cxxopts providing this query for `std::optional` value of given name and type.

No need for an if-check of `count` or catching an exception.

- - -
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/421)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -1538,6 +1538,15 @@ CXXOPTS_DIAGNOSTIC_POP
     return CXXOPTS_RTTI_CAST<const values::standard_value<T>&>(*m_value).get();
   }
 
+#ifdef CXXOPTS_HAS_OPTIONAL
+  template <typename T>
+  std::optional<T>
+  as_optional() const
+  {
+    return as<T>();
+  }
+#endif
+
   private:
   void
   ensure_value(const std::shared_ptr<const OptionDetails>& details)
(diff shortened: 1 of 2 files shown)


jarro2783:
Thanks for your PR. I will take a look properly when I get a chance in a couple of days.

nigels-com:
No rush.  We're having some interesting internal discussion.
Want to check a few assumptions when I get a chance.

include/cxxopts.hpp
```diff
@@ -1538,6 +1538,22 @@ CXXOPTS_DIAGNOSTIC_POP
...
   }
 
+#ifdef CXXOPTS_HAS_OPTIONAL
+  template <typename T>
+  std::optional<T>
+  optional() const
+  {
+    try
+    {
+      return as<T>();
```
nigels-com, line 1548: Our thought is that cxx still ought to throw for type mismatch (asking for int as a string, day), and return `std::nullopt` if the name lookup fails (undefined value, missing CLI parameter).

jarro2783: That seems like user error though, since you should know what types you created your options with. In which case it should still throw.

jarro2783:
This looks good, I'm also thinking about whether there is a better name for the function. I wonder if there some convention from other projects like something with `maybe` in the name.

include/cxxopts.hpp
nigels-com: Yes, agreed.  Will ammend.

jarro2783:
What do you think about calling it `maybe_as`, I think that fits the convention when a function can return an optional, and makes it clear that it is the same behaviour as `as`.

nigels-com:
I was thinking of `make_optional` in the manner of [std::make_shared](https://en.cppreference.com/w/cpp/memory/shared_ptr/make_shared).

It turns out that [std::make_optional](https://en.cppreference.com/w/cpp/utility/optional/make_optional) does already exist too.  (News to me!)

nigels-com:
I understand your reasoning with _maybe_ but isn't that just a synonym for optional?
Is clarity improved by assigning a "maybe" to an "optional"?

nigels-com:
Yes, `as_optional` sounds good to me.

nigels-com:
Updated as discussed.

PR #423, opened 2024-03-05
Fix a number of packaging issues

dcbaker:
This little series fixes or improves a number of package issues:
  - the pkg-config file has an empty Description: field
  - the pkg-config file now has it's `URL` field set
  - On NixOS, using CMAKE_INSTALL_DATAROOTDIR will put files in the proper arch-independent location
  - the pkg-config file now has it's `Requires` field and Cflags correctly configured when ICU is used
  - the CMake files and pkg-config files are properly installed in arch-specific locations when ICU is used. Since ICU is a arch specific the packaging files are also arch specific when used with ICU

- - -
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/423)


jarro2783:
Thanks, looks good to me.

PR #427, opened 2024-04-19
Add <limits> include to cxxopts.hpp

MausKaffee34767:
Build fails with `external/cxxopts/include/cxxopts.hpp:515:29: error: ‘numeric_limits’ is not a member of ‘std’`.
Full log:
wasmdec

I found the fix here: https://stackoverflow.com/questions/71296302/numeric-limits-is-not-a-member-of-std

- - -
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/427)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -42,6 +42,7 @@ THE SOFTWARE.
 #include <vector>
 #include <algorithm>
 #include <locale>
+#include <limits>
 
 #ifdef CXXOPTS_NO_EXCEPTIONS
 #include <iostream>


include/cxxopts.hpp
```diff
@@ -42,6 +42,7 @@ THE SOFTWARE.
 #include <vector>
 #include <algorithm>
 #include <locale>
+#include <limits>
```
jarro2783, line 45: this is a bit strange, because it is already included on line 33

MausKaffee34767:
Sorry, this is a bug in the wasmdec subrepo!

PR #428, opened 2024-04-22
Fix example and test compiled warnings and error in MSVC.

gwankyun:
Tested in VS2022.

- - -
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/428)


jarro2783:
Looks like the build options changed something because the unit tests are segfaulting now.

PR #437, opened 2024-07-27
Support external definition of CXXOPTS_ENABLE_INSTALL option

xukeawsl:
I hope to support the installation of cxxopts together with the main project


lazysegtree:
@jarro2783 Can you take a look at this PR? It's been pending for a long time.

jarro2783:
According to the docs, `option` does nothing if the variable is already set, so does this PR actually change any behaviour?

PR #444, opened 2024-12-02
Multitoken options

maickrau:
Implements multitoken options from Boost.Program_options. A multitoken option will take values until the next option is given. This makes it possible to switch to cxxopts without breaking backwards compatibility for multitoken options, and also enables a slightly easier syntax for the user for options with multiple values.

Example code without multitoken which just takes multiple parameters and prints them:
```
#include <iostream>
#include <vector>
#include <string>
#include "cxxopts/include/cxxopts.hpp"

int main(int argc, char** argv)
{
	cxxopts::Options options { "test" };
	options.add_options()
		("a", "Option a")
		("b", "Option b", cxxopts::value<std::vector<std::string>>())
		("c", "Option c", cxxopts::value<std::vector<std::string>>())
		;
	auto params = options.parse(argc, argv);
	bool a = params.count("a");
	std::cerr << "a: " << (a ? "yes" : "no") << std::endl;
	std::cerr << "b:";
	if (params.count("b") != 0)
	{
		for (std::string value : params["b"].as<std::vector<std::string>>())
		{
			std::cerr << " " << value;
		}
	}
	std::cerr << std::endl;
	std::cerr << "c:";
	if (params.count("c") != 0)
	{
		for (std::string value : params["c"].as<std::vector<std::string>>())
		{
			std::cerr << " " << value;
		}
	}
	std::cerr << std::endl;
	std::cerr << "unmatched:";
	for (std::string value : params.unmatched())
	{
		std::cerr << " " << value;
	}
	std::cerr << std::endl;
}
```

```
$ ./main -b 1 2 -c 3 4 -a 5
a: yes
b: 1
c: 3
unmatched: 2 4 5
```

Applying multitoken to option `-c` while keeping `-b` as is:
```
		("b", "Option b", cxxopts::value<std::vector<std::string>>())
		("c", "Option c", cxxopts::value<std::vector<std::string>>()->multitoken())
```

will now cause option `-c` to take values until the next option is given:

```
$ ./main -b 1 2 -c 3 4 -a 5
a: yes
b: 1
c: 3 4
unmatched: 2 5
```

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -386,6 +386,9 @@ class Value : public std::enable_shared_from_this<Value>
   is_container() const = 0;
 
   virtual bool
+  is_multitoken() const = 0;
+
+  virtual bool
   has_implicit() const = 0;
 
   virtual std::string
@@ -393,6 +396,9 @@ class Value : public std::enable_shared_from_this<Value>
 
   virtual std::string
   get_implicit_value() const = 0;
+
+  virtual std::shared_ptr<Value>
+  multitoken() = 0;
 
   virtual std::shared_ptr<Value>
   default_value(const std::string& value) = 0;
@@ -1189,6 +1195,12 @@ class abstract_value : public Value
     return type_is_container<T>::value;
   }
 
+  bool
+  is_multitoken() const override
+  {
+    return m_multitoken;
+  }
+
   void
   parse() const override
   {
@@ -1205,6 +1217,13 @@ class abstract_value : public Value
   has_implicit() const override
   {
     return m_implicit;
+  }
+
+  std::shared_ptr<Value>
+  multitoken() override
+  {
+    m_multitoken = true;
+    return shared_from_this();
   }
 
   std::shared_ptr<Value>
@@ -1264,6 +1283,7 @@ class abstract_value : public Value
 
   bool m_default = false;
   bool m_implicit = false;
+  bool m_multitoken = false;
 
   std::string m_default_value{};
   std::string m_implicit_value{};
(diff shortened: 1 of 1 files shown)


jarro2783:
I'll have to get back to you on this one, I haven't forgotten it, but I can't comment further right now.

jarro2783:
I wonder if it is worth putting some help text in to let users know that something is a multitoken option. Other than that this looks good.

patrikhuber:
Missing multitoken support is also the last bit preventing me from completely ditching boost program_options and switching to cxxopts fully. Would be really nice to have this.

I'm assuming this would also work with `cxxopts::value<std::vector<float>>()->multitoken()`?

maickrau:
@jarro2783 I was just going by the Boost implementation which doesn't change the help text for the parameter, but if you would like to add a help text that's fine by me since it shouldn't affect backwards compatibility.

@patrikhuber Yes, float works as well. Multitoken options use the same `parse_option` method as other options so I think it should work for all types.

jarro2783:
I'm not too fussed, I thought it might be good so that it is immediately obvious to the users of your program.
I'll start with this PR and we can always add to it later.

lazysegtree:
@maickrau, are you still tracking this? Will you be able to follow up on the review comments posted above?

jarro2783:
Did you still want to get this merged. Can you start with rebasing since the builds are now fixed up?

eyalroz:
@maickrau : Ping.

include/cxxopts.hpp
```diff
@@ -385,6 +385,9 @@ class Value : public std::enable_shared_from_this<Value>
   virtual bool
   is_container() const = 0;
 
+  virtual bool
```
eyalroz, line 388: Where in the code is the documentation (e.g. doxygen) of the multitoken option?

Issue #439, opened 2024-09-30
Path with spaces gets broken up at space


patrikhuber:
Hi,

I've got a simple app with an `--input` argument that contains a path with a space. I'm using VS and `launch.vs.json` to launch this, so the `args` part of `launch.vs.json` looks like:

```
      "args": [
        "--input",
        "\"C:\\My - Long path\""
      ]
```

And I'm reading this into cxxopts with `input_path = result["input"].as<std::filesystem::path>();`.

The problem is that `input_path` only contains `C:\\My`.

I've tried it with and without the escaped quotes in launch.vs.json (i.e. with and without the extra `\"` but it didn't make any change.
I can also confirm that `argv[2]` does confirm the whole path, i.e. argv[2] is a char* containing `"C:\\My - Long path"`.

Seems cxxopts is parsing / breaking this up wrongly?

jarro2783:
I suspect that because you have used `std::filesystem::path` it is getting put through the stringstream parser which is splitting it at the first space. If you make it a string then it should work. It should probably just work though, I'll see if I can work out a proper fix.

patrikhuber:
Thanks @jarro2783. Any chance to get a fix in for this? `std::filesystem::path` is pretty commonly used these days.

jarro2783:
Not yet, I'll try and get some time this week.

nigels-com:
@patrikhuber I had a go at supporting this, can you give that branch a try for your use case?  PR #447
Would appreciate any feedback.

lazysegtree:
@patrikhuber Were you able to test it? Are you still facing this issue?

PR #447, opened 2024-12-21
C++17 std::filesystem::path support added

nigels-com:
Main thing for supporting C++17 std::filesystem::path is the specialisation for `parse_value`.

Test coverage added.

```
$ test/options_test '[path]' 
Filters: [path]
===============================================================================
All tests passed (3 assertions in 1 test case)
```

For Issue #439

- - -
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/447)


jarro2783:
I think this works, but I suspect skipping whitespace in the stringstream parser is probably also correct since it makes sense to always parse the whole value.

PR #460, opened 2025-06-27
fix the issue with default empty vectors behaviour. update unit test

### LINKED ISSUE #459: Nonconsistent behavior of (default) empty vector arguments.
# Problem description

Specifying empty string as default argument for `vector<T>` argument type leads to different behavior based on `T`: `vector<string>` leads to vector of size 1 with an empty string argument, while for `vector<double>` I get `incorrect_argument_type`. Non of the examples seem to be expected behavior.

## Affected versions

Checked 3.2.0 and 3.3.1 -- both have this bug.

# Expected behavior

Empty string value for vector argument results in empty `vector<T>`.

# Minimalistic example

```cpp
#include "cxxopts.hpp"

#include <iostream>
#include <string>
#include <vector>

using namespace std;

int main(int argc, char* argv[])
{
  vector<double> dvec;
  vector<string> svec;

  cxxopts::Options options("default-vectors", "demostrates default empty vectors behaviour");
  options.add_options()
    ("dvec", "a vector of doubles", cxxopts::value(dvec)->default_value(""))
    ("svec", "a vector of strings", cxxopts::value(svec)->default_value(""))
    ("h,help", "display help and exit")
  ;
  try {
    auto res = options.parse(argc, argv);
    if (res.count("help") > 0) {
      cout << options.help() << endl;
      return 0;
    }

    cout << "dvec[" << dvec.size() << "]: ";
    for (const auto& v: dvec) {
      cout << v << ",";
    }
    cout << endl;
    cout << "svec[" << svec.size() << "]: ";
    for (const auto& v: svec) {
      cout << "'" << v << "',";
    }
    cout << endl;
  } catch (const cxxopts::exceptions::incorrect_argument_type &e) {
    cout << "parsing error: " << e.what() << endl;
    return 1;
  }
  return 0;
}
```

## Observed output

```bash
$ ./default-vectors --dvec=1
dvec[1]: 1,
svec[1]: '',

$ ./default-vectors --help
parsing error: Argument '' failed to parse
```

## Expected output

```bash
$ ./default-vectors --help
demostrates default empty vectors behaviour
Usage:
  default-vectors [OPTION...]

      --dvec arg  a vector of doubles (default: "")
      --svec arg  a vector of strings (default: "")
  -h, --help      display help and exit

$ ./default-vectors --dvec=1
dvec[1]: 1,
svec[0]:
```

# Suggested fix

I have found a bug in `parse_value(const std::string& text, std::vector<T>& value)`,  see suggested in PR #460

vicproon:
fixes #459

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -1100,22 +1100,19 @@ void parse_value(const std::string& text, char& c)
   c = text[0];
 }
 
+template<typename T> void add_value(const std::string& text, std::vector<T>& value);
+
 template <typename T>
 void
 parse_value(const std::string& text, std::vector<T>& value)
 {
   if (text.empty()) {
-    T v;
-    parse_value(text, v);
-    value.emplace_back(std::move(v));
     return;
   }
   std::stringstream in(text);
   std::string token;
   while(!in.eof() && std::getline(in, token, CXXOPTS_VECTOR_DELIMITER)) {
-    T v;
-    parse_value(token, v);
-    value.emplace_back(std::move(v));
+    add_value(token, value);
   }
 }
 
--- a/test/options.cpp
+++ b/test/options.cpp
@@ -709,6 +709,40 @@ TEST_CASE("std::vector", "[vector]") {
   CHECK(vector[3] == 4.5);
 }
 
+TEST_CASE("empty std::vector", "[vector]") {
+  std::vector<double> double_vector;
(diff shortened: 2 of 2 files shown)


vicproon:
@jarro2783 Hello, when will you have a spare minute to review the fix?

jarro2783:
Looks good, thanks.

Issue #463, opened 2025-09-05
When using gtest, I encountered a problem with competing app parameters(Problem with empty argc parameter)


hana-kami:
In cxxopts.hpp:2524

while (current != argc)
  {
    if (strcmp(argv[current], "--") == 0)
    {
      consume_remaining = true;
      ++current;
      break;
    }
......
}

Change it here to while (current < argc) to prevent this incoming parameter from being null

houstudio:
- options.parse(1,nullptr);options.parse(0,nullptr);will caused crash
- while(current != argc)change to while(current <argc)will more strong:)

jarro2783:
Is your argc zero?

hana-kami:
> Is your argc zero?

No, I have the parameters of gtest, but when gtest encounters --, it passes the subsequent parameters to the test case, but my application is also trying to parse these parameters

hana-kami:
I'm sorry if my explanation wasn't clear enough. The issue is that testing::InitGoogleTest(&argc, argv) removes processed parameters, but it's over-processing my parameters, resulting in argc = 0 and argv = 0x0 (null). When my application tries to parse these parameters, it encounters this problem:

> ```
> Thread 1 "gui_test" hit Breakpoint 1, cxxopts::OptionParser::parse (this=0x7fffffffb840, argc=0, argv=0x0) at cxxopts.h:2526
> 2526        if (strcmp(argv[current], "--") == 0)
> ```

Would it be better for compatibility to modify the code in cxxopts.hpp around line 2524 as follows? Other users might encounter similar usage scenarios:
In cxxopts.hpp:2524

```
while (current < argc)
{
    if (strcmp(argv[current], "--") == 0)
    ......
}

```

jarro2783:
Thanks for the explanation. I think it is fine to change that condition to `<`.

PR #465, opened 2025-09-16
Fixes potential segmentation fault in edge cases.

hana-kami:
Check for null pointer before accessing argv[current] in OptionParser::parse() to handle invalid argument scenarios that can occur with over-processed command line arguments.

- - -
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/465)


jarro2783:
Fixes #463

PR #476, opened 2025-12-19
Apply minor code improvements

fiesh:
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/476)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -2012,9 +2012,9 @@ class Options
     const std::string& group,
     const std::string& short_name,
     const std::string& single_long_name,
-    std::string desc,
+    const std::string& desc,
     const std::shared_ptr<const Value>& value,
-    std::string arg_help
+    const std::string& arg_help
   )
   {
     OptionNames long_names;


lazysegtree:
@fiesh Can you 
- Make sure that builds that were originally passing still pass.
- Add explanation of what you are changing, and why.

include/cxxopts.hpp
```diff
@@ -468,16 +463,16 @@ class invalid_option_format : public specification
 {
   public:
   explicit invalid_option_format(const std::string& format)
-  : specification("Invalid option format " + LQUOTE + format + RQUOTE)
+  : specification("Invalid option format '" + format + '\'')
```
lazysegtree: This hardcoding of the constants hampers maintainability, and should not be done.

jarro2783:
I would prefer to leave the quote symbols as they are. They are unlikely to be changed, but it is easier if they ever are changed, and it is still readable this way.

fiesh:
Among the reasons why I changed this:

* The `ALL_CAPS` naming is odd and typically only used for macros
* This creates an unnecessary `std::string` in every translation unit that includes the header.  Its ctor is non-trivial and may, depending on the implementation and whether it supports SSO, involve a memory allocation
* It may slow down program start-up
* The only reason why these appear to be `std::string` and not say a `constexpr char` seems to be that `std::string::operator+` is implicitly used later on.  Don't think that's a good reason to use a sub-optimal type.

This also leads to code linters like `clazy` rightfully complain, which brought up the issue to begin with.  But if the change isn't desired here, I'm happy to keep it a downstream patch that we do.

lazysegtree:
@fiesh If you just put the title in your PR as "minor code improvements" without explaining any reasoning behind your change, people are bound to misunderstand your intentions.

lazysegtree:
Now I see your intentions, and the issues in the code : unconventional naming, linter flags, extra allocations. This change, or some alternatives to it might make sense.

I would still like to see an alternative where we don't have to hardcode the constant.

lazysegtree:
I do plan to apply static code analyser/linter in this project later on, and then I might revisit this.

fiesh:
> @fiesh If you just put the title in your PR as "minor code improvements" without explaining any reasoning behind your change, people are bound to misunderstand your intentions.

This was the commit message:

[Remove superfluous non-pod globals](https://github.com/jarro2783/cxxopts/pull/476/commits/b969c2f5b638ce49a6342f393076495b1207d320)
These make everybody pay for their initialization upon program start and
are completely unnecessary.

If you only read the PR's title and not the commit messages, blame's not on me.

PR #481, opened 2026-01-12
feat: Switch to stable_partition to ensure that only first long options is printed in help

### LINKED ISSUE #363: Use first, not last, long name in usage message

lazysegtree:
Because we were using `partition`, not `stable_partition` there was no gurantee about preserving the ordering of long options.

Some users have reported that it s not deterministic which one of the long options is selected for priting in help menu.
For example - https://github.com/jarro2783/cxxopts/issues/363

- Adding a test case, which you can see fails. (https://github.com/jarro2783/cxxopts/actions/runs/20923706555/job/60115751379)
- Switched to `stable_partition` and then test passes.

Fixes #363

This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/481)

### DIFF
--- a/include/cxxopts.hpp
+++ b/include/cxxopts.hpp
@@ -2339,7 +2339,7 @@ OptionAdder::operator()
     // (length-1) and longer names
   std::string short_name {""};
   auto first_short_name_iter =
-    std::partition(option_names.begin(), option_names.end(),
+    std::stable_partition(option_names.begin(), option_names.end(),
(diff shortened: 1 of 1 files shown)


jarro2783:
I like this, thanks.

eyalroz:
One-liner patches fixing bugs - the best kind :-)

PR #498, opened 2026-05-21
Avoid undefined std::abs(INT_MIN) in integer_parser

rootvector2:
## Bug

`integer_parser` in `include/cxxopts.hpp` (line 1060 on master) computes
the upper bound of allowed magnitude for a negative parse via:

```cpp
limit = static_cast<US>(std::abs(static_cast<intmax_t>((std::numeric_limits<T>::min)())));
```

When `T` is a signed integer whose `min()` equals `INTMAX_MIN` (true for
`int64_t` / `long long` on every platform where `intmax_t` is 64 bits,
which is the common case), the cast to `intmax_t` is exact and the
subsequent `std::abs(INTMAX_MIN)` is **undefined behaviour**: negating
the most-negative two's-complement value cannot be represented in the
signed type. cppreference and POSIX both document this for `abs/llabs`.

This is triggered on every parse of a negative `int64_t`, not only at
the boundary. The minimal trigger is `integer_parser<int64_t>(\"-1\", v)`.

## Reproduction

```cpp
// repro.cpp
#include \"cxxopts.hpp\"
#include <cstdint>
int main() {
    int64_t v;
    cxxopts::values::integer_parser<int64_t>(\"-1\", v);
}
```

```
$ clang++ -std=c++17 -fsanitize=undefined -fno-sanitize-recover=all \
      -I include repro.cpp -o repro && ./repro
stdlib.h:111:101: runtime error: negation of -9223372036854775808 cannot
be represented in type 'long'; cast to an unsigned type to negate this
value to itself
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior
```

## Fix

Compute `|min|` directly in unsigned arithmetic. For two's-complement
signed types, `|min| == max + 1`, and that value is representable in the
unsigned counterpart `US`. The arithmetic stays inside `US`, so no
intermediate signed overflow occurs.

```cpp
limit = static_cast<US>(static_cast<US>((std::numeric_limits<T>::max)()) + US{1});
```

The fix is local to the callee. No caller has to change.

For unsigned `T` (where `negative == true` will be rejected later by
`checked_negate`), the new expression wraps to `0`, matching the old
`std::abs(0)` value — so behaviour for the unsigned path is preserved.

## Evidence

- UBSan run before the patch: aborts on `integer_parser<int64_t>(\"-1\", v)`
  with the diagnostic above.
- UBSan run after the patch: clean, prints `-1`.
- Boundary cases exercised after the patch (all match expected results,
  no UBSan reports): `int64_t` parses of `-9223372036854775808` (= INT64_MIN),
  `9223372036854775807` (= INT64_MAX), `-9223372036854775809` (rejected),
  `9223372036854775808` (rejected); analogous cases for `int32_t`, `int8_t`,
  `uint64_t`; and hex forms `-0x8000000000000000`, `0x7fffffffffffffff`,
  `-0x8000000000000001` (rejected).
- Existing test suite (`ctest` against the in-tree CMake build, both the
  regex and `CXXOPTS_NO_REGEX` configurations) passes:
  `100% tests passed, 0 tests failed out of 4`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/498)


lazysegtree (approved):
Thanks. Looks correct.

Optional improvement suggestion - Add new test that will fail with older code ( test for int64_t with UB sanitizers in CI test arguments)

Example - https://github.com/lazysegtree/cxxopts/pull/7/changes (First commit that just add tests causes failures. Second commits makes the CI green)

rootvector2:
Good call. Added the INT64_MIN parse to the boundary test and turned on UBSan (-fsanitize=undefined -fno-sanitize-recover=undefined) for options_test/options_test_noregex under Clang/GCC. Verified it aborts on the old std::abs line and passes with the fix, so it'll catch a regression.

jarro2783:
The windows build fails, so maybe that compiler check is not correct.

lazysegtree:
Sorry, I missed the Windows failure while submitting the suggestion about the CI change. @rootvector2 You may add NOT WIN32 in cmake. All builds then pass as seen in [lazysegtree#7](https://github.com/lazysegtree/cxxopts/pull/7) after the latest commit.

rootvector2:
Done — added NOT WIN32 to the guard so the UBSan flags only get applied on Clang/GCC where they actually work. clang-cl was matching the Clang branch but doesn't take -fsanitize=undefined that way.

jarro2783:
Your PR was very detailed, thanks for your contribution.

PR #499, opened 2026-05-29
Fix parsing special floating point values

### LINKED ISSUE #314: Handle "inf", "-inf" and "nan" for double and float types
Currently when parsing values as `double` or `float`, cxxopts  uses the fallback parser using stringstream. However, using input streams does not support a representation for infinity and `NaN` (see [this stack overflow question](https://stackoverflow.com/questions/11420263/is-it-possible-to-read-infinity-or-nan-values-using-input-streams)) .

It seems that all that is needed to support passing "infinity" and similar values that can be properly parsed is to add two  specialization of `parse_value`, one for `double`, that uses `std::stod`, and one for `float`, that uses `std::stof`.

Another specialization for `long double` could also be added that uses `std::stold`.

Note that `std::stod`, `std::stof` and `std::stold` were [introduced in C++11](https://en.cppreference.com/w/cpp/string/basic_string/stof).

lilong555:
Fixes #314.

This adds explicit parsing for `float`, `double`, and `long double` using the C `strtof` / `strtod` / `strtold` APIs instead of the generic stream parser. That allows values such as `inf`, `infinity`, `-inf`, and `nan` to parse while keeping support for `CXXOPTS_NO_EXCEPTIONS` builds.

Validation run locally:
- `g++ -std=c++11 -Iinclude -Itest test/main.cpp test/options.cpp -o options_test.exe && .\options_test.exe`
- `g++ -std=c++11 -DCXXOPTS_NO_REGEX -Iinclude -Itest test/main.cpp test/options.cpp -o options_test_noregex.exe && .\options_test_noregex.exe`
- `g++ -std=c++11 -fno-exceptions -DCXXOPTS_NO_EXCEPTIONS -Iinclude -x c++ - -o noexceptions_smoke.exe`
- `g++ -std=c++11 -fno-exceptions -DCXXOPTS_NO_EXCEPTIONS -DCXXOPTS_NO_REGEX -Iinclude -x c++ - -o noexceptions_noregex_smoke.exe`

- - -
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/499)


jarro2783:
Thanks, I think this looks good for the default behaviour.

PR #500, opened 2026-09-03
Issue 489 : Allow disabling vector delimiter parsing per option

### LINKED ISSUE #489: Allow disabling vector delimiter based parsing per option
Use-case -> Two options `a` and `b` of value type `vector`, for one of them we want to allow delimiter based parsing, and for one, we don't.

`-a 1,2 -b 2,3` should parse as one value `"1,2"` for `a` and two values `"2"` and `"3"` for `b`.

Something like `cxxopts::value<std::vector<std::string>>()->disable_vector_delimit()`

Has been brought up in these issues. So I think its worth implementing.
- #228 
- #262

Shray-glitch:
## Summary

Adds a per-option way to disable delimiter-based parsing for `std::vector<T>` values.

By default, vector parsing continues to use `CXXOPTS_VECTOR_DELIMITER` as before. 

Calling:

```cpp
cxxopts::value<std::vector<std::string>>()
  ->disable_list_delimiter()
```

causes each provided argument to be treated as a single vector element instead of splitting it on the delimiter.

This is useful for repeated options whose individual values may themselves contain commas.

## Example

```text
--location paris --location new,jersey
```

With delimiter parsing disabled, this produces:

```cpp
{"paris", "new,jersey"}
```

instead of:

```cpp
{"paris", "new", "jersey"}
```

## Tests

Added coverage for:

* disabling delimiter parsing for one option while preserving normal behavior for another
* repeated options containing delimiters
* default vector values with delimiter parsing disabled

All existing tests pass.

Fixes #489

- - -
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/500)


include/cxxopts.hpp
```diff
@@ -1237,6 +1240,39 @@ parse_value(const std::string& text, std::vector<T>& value)
   }
 }
 
+template <typename T>
+void
+parse_value(
```
eyalroz: Perhaps a less generic name for this function?

include/cxxopts.hpp
```diff
@@ -1237,6 +1240,39 @@ parse_value(const std::string& text, std::vector<T>& value)
...
+{
+  parse_value(text, value);
+}
+
+template <typename T>
+void
+parse_value(
+  const std::string& text,
+  std::vector<T>& value,
+  bool parse_vector_delimiter
```
eyalroz: It's a delimiter of a list of values on the command-line. The fact that we use an std::vector to store them does not mean the delimiter is associated with vectors. IMHO anyway.

README.md
```diff
@@ -240,6 +240,29 @@ If the list of values is quoted, then spaces can be included. For example:
...
+~~~
+--location paris --location new,jersey
+~~~
+
+With delimiter parsing disabled, this is parsed as: 
+
+```cpp
+std::vector<std::string>{
+  "paris",
+  "new,jersey"
```
eyalroz: Would try to make this example less artificial, since "new,jersey" seems silly.

include/cxxopts.hpp
Shray-glitch: Renamed it to parse_value_with_list_delimiter() to make its purpose more explicit.

include/cxxopts.hpp
Shray-glitch: Agreed. I changed the new API and internal naming to use “list delimiter” instead of “vector delimiter”.

README.md
Shray-glitch: Replaced it with the --buffer-dimensions example from #489, which better reflects the actual use case.

PR #501, opened 2026-09-07
test fallback parser with meson and bazel

aouxwoux:
runs the options suite with regex disabled in meson and bazel.

tested with meson 0.64.1 and gcc 15.2 on windows, icu disabled: all 3 tests pass. bazel resolves all 3 targets locally; its test runner needs c++ build tools on this machine.

refs #486.

- - -
This change is [Reviewable](https://reviewable.io/reviews/jarro2783/cxxopts/501)


jarro2783:
thanks. But it looks like the fuzz test now fails.

